PR #4, opened 2023-06-20
✏️

thibaultleouay:


turbo.json
```diff
@@ -4,7 +4,19 @@
...
+        "RESEND_API_KEY",
+        "!NEXT_PUBLIC_GIT_*",
+        "DATABASE_URL",
+        "TINY_BIRD_API_KEY",
+        "CLERK_SECRET_KEY",
+        "NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY",
+        "NEXT_PUBLIC_CLERK_SIGN_IN_URL",
+        "NEXT_PUBLIC_CLERK_SIGN_UP_URL",
+        "NEXT_PUBLIC_CLERK_AFTER_SIGN_IN_URL",
+        "NEXT_PUBLIC_CLERK_AFTER_SIGN_UP_URL"
```
mxkaske, line 18: Did you get an error on building and it solved the issue? Asking because I thought `*` would be enough and I got an error with the `RESEND_API_KEY`.

apps/web/src/app/action.ts
```diff
@@ -50,7 +51,6 @@ const send = async (email) => {
     from: "onboarding@openstatus.dev",
     to: "maximilian@kaske.org",
     subject: "Hello world",
-    // @ts-ignore FIXME:
```
mxkaske, line 54: Thanks!

PR #5, opened 2023-06-21
Feat/cron ping self monitor

mxkaske:
We schedule a cron job every 10 minutes to our new `/api/v0/ping` API route inside of the `vercel.json` file. 

> The cron jobs are [only invoked on production](https://vercel.com/guides/how-to-setup-cron-jobs-on-vercel#deploy-your-project). In the meantime, you can manually send requests to `/api/v0/ping` to monitor the current page.

```
{
  "crons": [
    {
      "path": "/api/v0/cron",
      "schedule": "*/10 * * * *"
    }
  ]
}
```

To allow server errors `500`, we have added a randomised value `const RANDOM = Math.random() > 0.9`.

We started using the `packages/tinybird` for granularity. Also `packages/upstash` has been added while currently not in use. We use `tsup` to compile the packages.


apps/web/src/app/page.tsx
```diff
@@ -1,18 +1,28 @@
...
+  { id: "openstatus", timestamp: 1687370131709, statusCode: 500 },
+  { id: "openstatus", timestamp: 1687370254931, statusCode: 200 },
+  { id: "openstatus", timestamp: 1687370255071, statusCode: 500 },
+];
 
-export const metadata: Metadata = {
-  title: "Open-source monitoring service",
-  description:
-    "OpenStatus is an open source alternative to your current monitoring service with beautiful status page",
-};
```
mxkaske, line 10: Moved it to `layout.tsx` which will then be the default for every subpage.

apps/web/src/app/api/v0/ping/route.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export async function GET(req: Request) {
+  const RANDOM = Math.random() > 0.5;
+  try {
+    // REMINDER: connect tinybird, upstash and planetscale
+    await redis.ping();
+    const res = new Response("OK", { status: 200 });
+    await monitor(200);
+    if (RANDOM) {
```
thibaultleouay, line 10: I dont get it ?

apps/web/src/app/layout.tsx
```diff
@@ -14,21 +14,25 @@ const calSans = LocalFont({
   variable: "--font-calsans",
 });
 
+const TITLE = "Open-source monitoring service";
```
thibaultleouay: For Title I would use the product name 
it would be better to have `Blog - OpenStatus`

apps/web/src/app/_components/background.tsx
```diff
@@ -11,7 +11,7 @@ export default function Background({
   useMouseMove();
   return (
     <>
-      <div className="absolute inset-0 z-[-1]">
+      <div className="fixed w-full h-full">
```
mxkaske, line 14: Fixing an issue on mobile/smaller screens.

apps/web/src/app/api/v0/cron/route.ts
```diff
@@ -0,0 +1,48 @@
...
+    const { searchParams } = new URL(req.url);
+    const hasUrl = searchParams.has("url");
+    const url = hasUrl ? searchParams.get("url") : `${DEFAULT_URL}/api/v0/ping`;
+
+    const startTime = Date.now();
+    const res = await fetch(url, { cache: "no-store" });
+    const endTime = Date.now();
+    const latency = endTime - startTime;
+
+    await monitor(res, { latency, url });
```
mxkaske, line 42: We allow `?url=` search params. If empty, we will use our `/api/v0/ping` as default. We should never `cache` the result and we store the latency in `ms`.

apps/web/src/app/api/v0/ping/route.ts
mxkaske: This is just to a random error for testing purposes. I updated the equation to have a .1% chance to fail.

packages/tinybird/src/client.ts
```diff
@@ -0,0 +1,43 @@
+import * as z from "zod";
+import type { Tinybird } from "@chronark/zod-bird";
```
mxkaske, line 2: Check out the [docs](https://github.com/chronark/zod-bird#ingesting-data). Perfect for use to have a tinybird schema validation.

apps/web/src/app/api/v0/cron/route.ts
thibaultleouay, line 42: We could do
```
Return await monitor...
```

Instead of creating two response

packages/tinybird/src/client.ts
thibaultleouay: It could be cool if we store the tiny bird pipe / schema in the repo too

thibaultleouay (approved):
Few comments and merge 🚀
Let's improve it after

thibaultleouay:
I don't understand why we need tsup for package ?

From turbo doc 
https://turbo.build/repo/docs/handbook/publishing-packages/bundling 

https://turbo.build/repo/docs/handbook/publishing-packages/bundling


I see most of the package as internal one but might be wrong

PR #6, opened 2023-06-23
🚧 wip database  schema

thibaultleouay:


packages/db/src/schema/status-job.ts
```diff
@@ -0,0 +1,42 @@
...
+  id: int("id").autoincrement().primaryKey(),
+  jobType: mysqlEnum("job_type", ["website", "cron", "other"])
+    .notNull()
+    .default("other"),
+  periodicity: mysqlEnum("periodicity", [
+    "every-5-min",
+    "every-1-min",
+    "every-1-h",
+    "other",
+  ])
```
mxkaske: Let's call the prop `frequency` and the values `1m`, `5m`, `30m` and `1h`

packages/db/drizzle/0000_overconfident_molten_man.sql
```diff
@@ -0,0 +1,45 @@
...
+	`updated_at` timestamp NOT NULL ON UPDATE CURRENT_TIMESTAMP);
+--> statement-breakpoint
+CREATE TABLE `incidentUpdate` (
+	`id` int AUTO_INCREMENT PRIMARY KEY NOT NULL,
+	`incident_date` datetime,
+	`title` varchar(256),
+	`message` text,
+	`incident_id` int NOT NULL,
+	`created_at` timestamp NOT NULL DEFAULT (now()),
+	`updated_at` timestamp NOT NULL ON UPDATE CURRENT_TIMESTAMP);
```
mxkaske: What is the purpose of the table?

We will have to store every single change of an incident. Could be interesting to use tinybird for that.

packages/db/src/schema/page.ts
```diff
@@ -0,0 +1,27 @@
...
+import { relations } from "drizzle-orm";
+import { incident } from "./incident";
+import { statusJob } from "./status-job";
+
+export const page = mysqlTable("page", {
+  id: int("id").autoincrement().primaryKey(),
+
+  workspaceId: int("workspace_id").notNull(),
+
+  slug: varchar("slug", { length: 256 }), // which is used for https://slug.openstatus.dev
```
mxkaske, line 11: We should store the `slug` inside a `workspace`. So that every company has only a single status page with all the websites.

packages/db/src/schema/status-job.ts
```diff
@@ -0,0 +1,42 @@
...
+  mysqlEnum,
+  mysqlTable,
+  int,
+  varchar,
+  timestamp,
+} from "drizzle-orm/mysql-core";
+import { relations } from "drizzle-orm";
+import { page } from "./page";
+
+export const statusJob = mysqlTable("statusJob", {
```
mxkaske: Hearing the name "status job" for the first time. Should we call it "ping"? I have something similar where I called it function `monitor`, but all are related to pinning a website.

packages/db/drizzle/0000_overconfident_molten_man.sql
thibaultleouay: An incident in my mind was a set message you post when your service is down! 
and the message is this table, an update to an incident

packages/db/src/schema/page.ts
thibaultleouay: I thought we could have multiple page for a workspace :)

packages/db/src/schema/status-job.ts
thibaultleouay: yep monitor makes sense!

PR #7, opened 2023-06-25
Init clerk webhook

thibaultleouay:


apps/web/src/app/api/ping/route.ts
mxkaske, line 1: Why not using the `/api/v0/ping` route?

apps/web/src/middleware.ts
```diff
@@ -1,7 +1,7 @@
 import { authMiddleware } from "@clerk/nextjs";
 
 export default authMiddleware({
-  publicRoutes: ["/", "/api/og"],
+  publicRoutes: ["/", "/api/og", "/api/webhook/ping", "/api/webhook/clerk"],
```
mxkaske: We will need to check for placeholders here. I’m doing the same currently

packages/api/src/router/clerk/type.ts
mxkaske, line 1: Nice!

mxkaske (commented):
🚀

apps/web/src/middleware.ts
thibaultleouay: I think we need to do with the matcher instead for the middle ware ?

PR #13, opened 2023-06-29
Update README.md

Adamkaram:
important to note that  were env file exist and what is required to make this project work


mxkaske:
Thanks a lot for the contribution! You raise a very good topic here. We aren't focusing on how to run it as for now but still, this should never miss in a README. 🚀

PR #14, opened 2023-07-01
Improve cron jobs

thibaultleouay:


apps/web/src/app/api/checker/cron/10m/route.ts
```diff
@@ -0,0 +1,9 @@
+import { cron } from "../_cron";
+
+export const runtime = "edge";
+export const preferredRegion = ["auto"];
```
mxkaske, line 4: Nice to have it here, even if default. Reminds us!

apps/web/src/app/api/checker/regions/_checker.ts
```diff
@@ -0,0 +1,77 @@
+import { Receiver } from "@upstash/qstash/cloudflare";
```
mxkaske, line 1: What's the benefit over using `@upstash/qstash/nextjs`?

apps/web/src/app/api/checker/regions/arn1/route.ts
```diff
@@ -0,0 +1,9 @@
+import { checker } from "../_checker";
+
+export const runtime = "edge";
+export const preferredRegion = ["arn1"];
+export const dynamic = "force-dynamic";
+
+export async function POST(request: Request) {
```
mxkaske, line 7: We could move it to a `GET` method and allow the `url` to be a search param.

packages/tinybird/src/client.ts
```diff
@@ -4,18 +4,24 @@ import * as z from "zod";
...
+  workspaceId: z.string(),
+  pageId: z.string(),
+  monitorId: z.string(),
+  timestamp: z.number().int(),
+  statusCode: z.number().int(),
+  latency: z.number().int(), // in ms
+  url: z.string(),
+  metadata: z.string().optional().default(""),
+  region: z.string().min(4).max(4), // TODO: object + stringify json
+});
```
mxkaske, line 18: We will probably have two different schemas for the `publishPingResponse` (datasource) and `getResponseList` (pipe). 

We don't have to return all the values, could group the entries,.. or transform the values e.g.

```
metadata: z.string().transform((m) => JSON.parse(m))
```
while saving it as string.

mxkaske (commented):
Good stuff, man!

apps/web/src/app/api/checker/regions/arn1/route.ts
mxkaske: Somehow, I think `POST` is cleaner. Having the props inside the `request.body` is cleaner.

apps/web/src/app/api/checker/regions/_checker.ts
thibaultleouay: We are in the edge environment which is the same as cloudfare worker 

We cant use all the node api

apps/web/src/app/api/checker/regions/arn1/route.ts
thibaultleouay: Qstash post a Json to an endpoint :)

packages/tinybird/src/client.ts
thibaultleouay: Sure just extract it when I did some small update

mxkaske (approved):

PR #22, opened 2023-07-06
Aggregated monitor data and search params as data table filter

### LINKED ISSUE #21: Tooltips don't work on mobile
On status page tooltip don't work on mobile

mxkaske:
Created a new tb pipe. It basically `GROUP BY cronTimestamp` and returns aggregated data like:

- average latency
- total number of entries
- number of entries being `ok` (status code between 200 and 299)

Each bar represents a full cron job execution, calling all the regions.

Additionally, we support `searchParams` to a specific degree and filter our datasource by `?cronTimestamp=123456789`. Connecting both allows us to open a details view for each cron job.

Question: When to differenciate between Operational, Downtime, Degraded?


apps/web/src/app/monitor/[id]/page.tsx
```diff
@@ -1,8 +1,33 @@
...
+  region: z.enum(availableRegions).optional(),
+  cronTimestamp: z.coerce.number().optional(),
+});
+
+export default async function Monitor({
+  params,
+  searchParams,
+}: {
+  params: { id: string };
+  searchParams: { [key: string]: string | string[] | undefined };
```
Comment (thibaultleouay) [inline apps/web/src/app/monitor/[id]/page.tsx:25]: We can use z.infer<typeof paramsschema> ?

Comment (mxkaske) [reply apps/web/src/app/monitor/[id]/page.tsx:25]: We shouldn't trust users input. Anywho can modify them. `?hello=world` would bypass it. That's why we validate schema on SSR.

apps/web/src/components/monitor/tracker.tsx
```diff
@@ -75,18 +76,23 @@ export function Tracker({ data, maxSize = 35 }: TrackerProps) {
...
                   <div className="flex justify-between">
                     <p className="text-xs font-light">
-                      {formatDistance(new Date(timestamp), new Date(), {
+                      {formatDistance(new Date(cronTimestamp), new Date(), {
                         addSuffix: true,
                         includeSeconds: true,
                       })}
                     </p>
-                    <p className="text-muted-foreground text-xs">{latency}ms</p>
+                    <p className="text-muted-foreground font-mono text-xs">
```
thibaultleouay: Maybe we should display it's the avg

PR #25, opened 2023-07-08
🚀 more api

thibaultleouay:


packages/db/src/schema/monitor.ts
```diff
@@ -16,7 +16,14 @@ export const monitor = mysqlTable("monitor", {
...
     .default("other")
     .notNull(),
-  frequency: mysqlEnum("periodicity", ["1m", "5m", "10m", "30m", "1h", "other"])
+  periodicity: mysqlEnum("periodicity", [
+    "1m",
+    "5m",
+    "10m",
+    "30m",
+    "1h",
+    "other",
```
mxkaske, line 25: How about removing "other"? Not very meaningful.

packages/db/src/schema/page.ts
```diff
@@ -35,3 +36,9 @@ export const pageRelations = relations(page, ({ many, one }) => ({
   }),
   monitors: many(monitor),
 }));
+
+// Schema for inserting a Page - can be used to validate API requests
```
mxkaske, line 40: We could use JSDoc comments. Just a side note 😃.

```js
/*
 *  Schema for inserting a Page - can be used to validate API requests
*/

```

mxkaske (approved):

PR #27, opened 2023-07-09
feat: more tinybird

mxkaske:
What to expect:

- new aggregated data via `groupBy` search param allowing to group the tinybird data for a whole day instead of for each cron job. For now, can be switched though the `ToggleButton`. Refactored `start` and `end` to `fromDate` and `toDate` and added them to `/monitor/[id]` path
- simplified the placeholder tracking bars
- add latest shadcn `Button`
- add simple `BackButton`
- new `.pipe` and `.datasource` files to include schemas in source code


apps/web/src/app/api/checker/cron/_cron.ts
```diff
@@ -46,6 +46,7 @@ export const cron = async ({
       url: `${DEFAULT_URL}/api/checker/regions/${region}`,
       body: {
         url: `${DEFAULT_URL}/api/ping`,
+        // FIXME: what is the concret timezone format? In DE, is 02h later
```
mxkaske, line 49: If you have any ideas?

thibaultleouay, line 49: Date.now return time since epoch in UTC

thibaultleouay (approved):
🚀

PR #42, opened 2023-07-11
🚀 monitor

thibaultleouay:
An example how to fetch the data  for the different pages 🚀

### DIFF
--- a/apps/web/src/app/app/(dashboard)/[workspaceId]/monitor/page.tsx
+++ b/apps/web/src/app/app/(dashboard)/[workspaceId]/monitor/page.tsx
@@ -2,10 +2,18 @@ import * as React from "react";
 
 import { Container } from "@/components/dashboard/container";
 import { Header } from "@/components/dashboard/header";
-import { wait } from "@/lib/utils";
+import { api } from "@/trpc/server";
 
-export default async function MonitorPage() {
-  await wait(1000);
+export default async function MonitorPage({
+  params,
+}: {
+  params: { workspaceId: string };
+}) {
+  const monitors = await api.monitor.getMonitorsByWorkspace.query({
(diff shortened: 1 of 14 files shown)


apps/web/src/middleware.ts
```diff
@@ -37,14 +42,19 @@ export default authMiddleware({
...
         .execute();
+
       if (result.length) {
         const orgSelection = new URL(
           `/app/${result[0].users_to_workspaces.workspaceId}`,
           req.url,
         );
         return NextResponse.redirect(orgSelection);
+      } else {
+        // return NextResponse.redirect(new URL("/app/onboarding", req.url));
```
thibaultleouay, line 55: @mxkaske that's where we should probably do something when the webhook is not received yet. 

redirect to a small onboarding page?

apps/web/src/app/app/(dashboard)/[workspaceId]/monitor/page.tsx
```diff
@@ -2,10 +2,18 @@ import * as React from "react";
...
 
-export default async function MonitorPage() {
-  await wait(1000);
+export default async function MonitorPage({
+  params,
+}: {
+  params: { workspaceId: string };
+}) {
+  const monitors = await api.monitor.getMonitorsByWorkspace.query({
+    workspaceId: Number(params.workspaceId),
```
Comment (thibaultleouay) [inline apps/web/src/app/app/(dashboard)/[workspaceId]/monitor/page.tsx:13]: Here we fetch the monitors 🚀
Some logic for all the components

apps/web/src/middleware.ts
mxkaske: Let's see what we can do. I don't mind having a Loading screen with a small description what is happening. Being open about the status.

mxkaske (approved):
Lgtm!

PR #52, opened 2023-07-13
💌 send emails when user subscribe

thibaultleouay:


Ipriyankrajai:
@mxkaske Once this code is deployed, we need to add a sitemap URL to the google search console.

packages/emails/emails/waitingList.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react";
```
mxkaske: Let's keep the file convention of: "waiting-list.tsx".

PR #62, opened 2023-07-14
fix: tinybird improvements

thibaultleouay:
generate an id for every request we put into tiny bird 
update the tinybird query to get the monitor response for a single monitor


mxkaske (approved):
lgtm!

PR #68, opened 2023-07-15
feat: improve og image and status-page

mxkaske:
#### Subdomain Status Page

- add `loading.tsx` and `layout.tsx`
- add `context` to `Tracker` to remove links and "eye" (keeping it simple for now)

CleanShot 2023-07-15 at 16 24 16@2x

#### Open Graph Image

- add dynamic tracker to it only if "?siteId" search param is attached
- only using it for the `/play` path (will be added to the "status-page" of each user - tbd if adding it to every page and defaulting to "openstatus")

CleanShot 2023-07-16 at 11 27 01@2x


apps/web/src/app/api/og/route.tsx
```diff
@@ -9,24 +12,48 @@ const size = {
...
   const calSemiBoldData = await calSemiBold;
 
   const { searchParams } = new URL(req.url);
+
   const title = searchParams.has("title") ? searchParams.get("title") : TITLE;
   const description = searchParams.has("description")
     ? searchParams.get("description")
     : DESCRIPTION;
+  const siteId = searchParams.has("siteId")
+    ? searchParams.get("siteId")
```
thibaultleouay: I'm a bit confused by siteID shouldn't get pageID or monitorId ?

mxkaske: So in tinybird, we currently use `siteId` for the query. You are absolutely right: this should be `monitorId`. I'll change the search param to it. But we will need a refactor (and `version` change to update the params for tinybird - this should be a different PR).

PR #69, opened 2023-07-15
fix:api for actions

thibaultleouay:
🚀

### DIFF
--- a/apps/web/src/app/app/(dashboard)/[workspaceId]/monitors/_components/action-button.tsx
+++ b/apps/web/src/app/app/(dashboard)/[workspaceId]/monitors/_components/action-button.tsx
@@ -41,11 +41,7 @@ import { api } from "@/trpc/client";
 
 type Schema = z.infer<typeof insertMonitorSchema>;
 
-interface Props {
-  props: Schema;
-}
-
-export function ActionButton({ props }: Props) {
+export function ActionButton(props: Schema) {
   const router = useRouter();
   const [dialogOpen, setDialogOpen] = React.useState(false);
   const [alertOpen, setAlertOpen] = React.useState(false);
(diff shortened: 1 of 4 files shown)


apps/web/src/app/app/(dashboard)/[workspaceId]/monitors/_components/action-button.tsx
```diff
@@ -37,35 +37,32 @@ import {
...
-  workspaceId: number;
-  url: string;
-  name: string;
-  description: string;
+  props: Schema;
 }
 
-// TODO: add correct types
-export function ActionButton({ workspaceId, ...props }: Props) {
+export function ActionButton({ props }: Props) {
```
Comment (mxkaske) [inline apps/web/src/app/app/(dashboard)/[workspaceId]/monitors/_components/action-button.tsx]: Feels double the wording. I'd suggest to use:

```suggestion
export function ActionButton(props: Schema) {
```

And update the props accordingly

apps/web/src/app/app/(dashboard)/[workspaceId]/monitors/page.tsx
```diff
@@ -23,7 +23,7 @@ export default async function MonitorPage({
       </Header>
       {monitors.map((monitor, index) => (
         <Container key={index} title={monitor.name} description={monitor.url}>
-          <ActionButton {...{ ...monitor, workspaceId }} />
+          <ActionButton props={monitor} />
```
Comment (mxkaske) [inline apps/web/src/app/app/(dashboard)/[workspaceId]/monitors/page.tsx]: Will be probably be then:

```suggestion
          <ActionButton {...monitor} />
```

apps/web/src/components/forms/montitor-form.tsx
```diff
@@ -43,6 +43,7 @@ export function MonitorForm({ id, defaultValues, onSubmit }: Props) {
       url: defaultValues?.url || "",
       name: defaultValues?.name || "",
       description: defaultValues?.description || "",
+      periodicity: defaultValues?.periodicity || undefined,
```
mxkaske, line 46: How about adding `"10m"` as default?

thibaultleouay: but if we set 10m by default the combo box will always be 10m and we won't see the empty states for the combo box

mxkaske: Fine by me. Thought it would require one less operation for the user to do. We do have the "description" and "label" for explanation.

thibaultleouay: CleanShot 2023-07-16 at 13 18 15

PR #73, opened 2023-07-18
Allow user to choose theme for status page

miljan-code:


apps/web/src/app/layout.tsx
```diff
@@ -47,12 +48,14 @@ export default function RootLayout({
   // If you want to develop locally with with Clerk,  Comment the provider below
   return (
     <ClerkProvider>
-      <html lang="en">
+      <html lang="en" suppressHydrationWarning>
```
mxkaske: Never seen before. Can we avoid suppressing warnings?

mxkaske (commented):
Thanks @miljan-code! Haven't seen that coming - but hey a page is not complete until it supports dark mode. 

Nothing to add on besides the comment! 🚀

apps/web/src/app/layout.tsx
miljan-code: Next-themes Theme Provider component causes hydration mismatches. So one of the solutions is to add suppressHydrationWarning to the html element, as said in the [docs](https://nextjs.org/docs/messages/react-hydration-error).

We can avoid suppressing warnings by using solution number 2 from the [docs](https://nextjs.org/docs/messages/react-hydration-error#solution-2-disabling-ssr-on-specific-components). That means that we need to use dynamic import and disable prerendering.

```typescript
import dynamic from "next/dynamic";

const ThemeProvider = dynamic(() => import("../components/theme-provider"), {
  ssr: false,
});
```

I can make a change if you want to avoid suppressing warnings.

mxkaske: Legend. Thanks for the explanation! Nothing to change here.

mxkaske:
@miljan-code, I've rebased and fixed the colors that are not been properly set. Gonna push the pr over the line. Only last bit to be added is the correct svg logos to support dark/light mode. Thanks again for your work done here.

PR #75, opened 2023-07-19
Turso experimentation

thibaultleouay:


packages/api/src/router/incident.ts
```diff
@@ -15,13 +15,17 @@ export const incidentRouter = createTRPCRouter({
   createIncident: protectedProcedure
     .input(insertIncidentSchema)
     .mutation(async (opts) => {
-      await opts.ctx.db.insert(incident).values(opts.input).execute();
```
mxkaske, line 18: If we don't use `await`, we can get rid of `async`. Seen in multiple places.

mxkaske (approved):
lfg

packages/db/drizzle/0000_loud_alice.sql
```diff
@@ -0,0 +1,71 @@
...
+	FOREIGN KEY (`workspace_id`) REFERENCES `workspace`(`id`) ON UPDATE no action ON DELETE no action
+);
+--> statement-breakpoint
+CREATE TABLE `user` (
+	`id` integer PRIMARY KEY NOT NULL,
+	`tenant_id` text(256),
+	`updated_at` integer DEFAULT (strftime('%s', 'now'))
+);
+--> statement-breakpoint
+CREATE TABLE `users_to_workspaces` (
```
mxkaske: I guess this is autogenerated. But doesn't fit well into the current naming convention (e.g. `incidentUpdate`).

PR #77, opened 2023-07-22
feat: add form error if slug not unique

mxkaske:


thibaultleouay:
Can we use use-debounce to avoid exhausting the back end 😁

PR #82, opened 2023-07-23
fix: monitor to page relation including form defaults

mxkaske:
- fix default values for the `defaultValue?.monitors` checkboxes
- fix `NaN` and add `EmptyState` to status page
- add `{where: monitors: {} }` relational query for the `getPagesByWorkspaces`
- fix all possible delete/update relations for the many-to-many relation `monitorsToPages`

### DIFF
--- a/packages/api/src/router/page.ts
+++ b/packages/api/src/router/page.ts
@@ -67,26 +67,57 @@ export const pageRouter = createTRPCRouter({
     .mutation(async (opts) => {
       const { monitors, ...pageInput } = opts.input;
 
-      await opts.ctx.db
+      const currentPage = await opts.ctx.db
         .update(page)
         .set(pageInput)
         .where(eq(page.id, Number(opts.input.id)))
         .returning()
         .get();
 
-      if (monitors && monitors.length) {
-        // We should make sure the user has access to the monitors
-        const values = monitors.map((monitorId) => ({
+      // TODO: optimize!
+      const currentMonitorsToPages = await opts.ctx.db
+        .select()
+        .from(monitorsToPages)
+        .where(eq(monitorsToPages.pageId, currentPage.id))
+        .all();
+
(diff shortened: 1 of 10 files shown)


packages/db/src/schema/page.ts
```diff
@@ -29,7 +29,7 @@ export const page = sqliteTable("page", {
 });
 
 export const pageRelations = relations(page, ({ many, one }) => ({
-  monitorsToPages: many(monitorsToPages),
+  monitors: many(monitorsToPages),
```
mxkaske: I'd suggest to use `monitors`. This fits well for the relational queries from drizzle. Than, we don't have to access the values via `page.monitorsToPages`.

packages/api/src/router/page.ts
```diff
@@ -129,16 +130,19 @@ export const pageRouter = createTRPCRouter({
...
-          ),
-        )
-        .all();
+      return opts.ctx.db.query.page.findMany({
+        where: and(
+          eq(page.workspaceId, opts.input.workspaceId),
+          inArray(page.workspaceId, workspaceIds),
+        ),
+        with: {
+          monitors: {
```
thibaultleouay: Does it really work because the relation is on pagesToMonitors ?

mxkaske: I renamed `monitorsToPages` to `monitors`. Just to use monitors all over the frontend. tbd.

packages/db/src/schema/page.ts
thibaultleouay: but it's a many-to-many relation ship we need a join table

thibaultleouay: https://orm.drizzle.team/docs/rqb#many-to-many 😭

mxkaske: Thanks a lot! Reverted back! 🚀

PR #87, opened 2023-07-26
feat: real data

thibaultleouay:


apps/web/src/app/api/checker/cron/_cron.ts
```diff
@@ -21,32 +27,54 @@ export const cron = async ({
...
+      await c.publishJSON({
+        url: `${DEFAULT_URL}/api/checker/region/${region}`,
+        body: payload,
+      });
+    }
+  }
+  if (periodicity === "10m") {
+    // Right now we are just checking the ping endpoint
+    for (const region of availableRegions) {
+      const payload: z.infer<typeof payloadSchema> = {
```
thibaultleouay, line 63: That's legacy

apps/web/src/app/api/checker/regions/_checker.ts
```diff
@@ -18,27 +21,42 @@ const tb = new Tinybird({ token: env.TINY_BIRD_API_KEY });
...
+        id: nanoid(), // TBD: we don't need it
+        workspaceId: monitorInfo.workspaceId,
+        pageId: pageId,
+        monitorId: monitorInfo.monitorId,
+        timestamp: Date.now(),
+        statusCode: res.status,
+        latency,
+        url: monitorInfo.url,
+        region,
+        cronTimestamp: monitorInfo.cronTimestamp || Date.now(), // TBC: why did we have this?
```
mxkaske: We don't need the default option. We had it because it was possibly null.

Instead of adding each property from `payloadSchema` we can deconstruct and add the props via `...monitorInfo`. Need to check regarding `pageIds` not to pass it then.

apps/web/src/app/api/checker/schema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const payloadSchema = z.object({
+  workspaceId: z.string(),
+  monitorId: z.string(),
+  url: z.string(),
+  cronTimestamp: z.number(),
+  pageId: z.array(z.string()),
```
mxkaske: Lets call it `pageIds` to make it explicit it's an array of ids.

apps/web/src/app/api/checker/cron/_cron.ts
```diff
@@ -21,32 +27,54 @@ export const cron = async ({
...
-
-  // Right now we are just checking the ping endpoint
-  for (const region of availableRegions) {
-    await c.publishJSON({
-      url: `${DEFAULT_URL}/api/checker/regions/${region}`,
-      body: {
+  const result = await db
+    .select()
+    .from(monitor)
+    .where(eq(monitor.periodicity, periodicity))
```
mxkaske: Also need to check if the monitor is active!

PR #103, opened 2023-07-29
fix: og monitor id

mxkaske:
fixes the og for the subdomains. We take the first monitor that we get. No filter, no sorting. This is a temporary fix.


thibaultleouay (approved):
🚀

PR #104, opened 2023-07-29
fix: public post folder for image

mxkaske:
Important to know: public should be in [root directory](https://nextjs.org/docs/pages/building-your-application/configuring/src-directory)


mxkaske:
A hotfix changing `Image` to `img` is out for now...

PR #106, opened 2023-07-30
🚀 add tracking

thibaultleouay:


packages/api/src/router/clerk/webhook.ts
```diff
@@ -47,11 +48,20 @@ export const webhookRouter = createTRPCRouter({
         .returning()
         .get();
 
-      await sendEmail({
```
mxkaske: Let's keep it?!

thibaultleouay:
I'll keep it as open until this is fixed : 
https://github.com/rudderlabs/rudder-sdk-node/issues/110

PR #131, opened 2023-08-01
post: first 48h being public

mxkaske:


nkilm:
Blog looks great!🔥 

Just a suggestion, can we center the tweets included in the blog? We can wrap the `Tweet` component with a wrapper to adjust the alignment and export in `mdx-components`

image

PR #133, opened 2023-08-02
hide tooltip if no description is provided

cskiwi:


mxkaske:
Thanks a lot @cskiwi! As simple as that!

Issue #144, opened 2023-08-02
Set up documentation site 


thibaultleouay:
Create a new app with nextra with the doc 
How to contribute to the project 
How to set up env !

thibaultleouay:
This has been shipped 🛳️

PR #147, opened 2023-08-03
Add documentation page using nextra

nkilm:
## Related Issues
#144 

## Changes
- New documentation page using [Nextra](https://nextra.site/)

I've set up a basic documentation site using the guide mentioned the official nextra site and starter template. We can improve upon this. Nextra covers most of the usecases i.e, syntax highlighting, cmd + k, table of contents, etc.

## References
- https://nextra.site/docs/docs-theme
- https://github.com/shuding/nextra-docs-template/

Any suggestions are welcome!

## Testing locally
cd into `/app/docs` and install using `pnpm i`. Later start the server using `pnpm dev`. It starts the server on PORT 5000.

## Visuals

image

image

### DIFF
--- a/apps/docs/.gitignore
+++ b/apps/docs/.gitignore
@@ -1,0 +1,37 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
+
+# testing
+/coverage
+
+# next.js
+/.next/
+/out/
+
+# production
+/build
+
+# misc
+.DS_Store
+*.pem
+
+# debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# local env files
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+# vercel
+.vercel
+
+# contentlayer
+.contentlayer
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -12,13 +12,14 @@
   "homepage": "https://openstatus.dev",
   "dependencies": {
     "next": "^13.4.12",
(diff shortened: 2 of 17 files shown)


apps/docs/.gitignore
thibaultleouay, line 1: can't we use the main .gitignore ?

apps/docs/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "start": "next start"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/openstatusHQ/openstatus"
+  },
+  "homepage": "https://openstatus.dev",
+  "dependencies": {
+    "next": "^13.4.12",
+    "nextra": "latest",
```
thibaultleouay: instead of latest can we specify the version to avoid some issue if there's an update ?

apps/docs/.gitignore
nkilm: yes we can. I thought it was kind of like a separate project so I left it like that.

apps/docs/package.json
nkilm: Alright.

apps/docs/tsconfig.json
thibaultleouay, line 1: we should also extend our own tsconfig 

in dev dependencies add "tsconfig":"workspace:*"

apps/docs/package.json
nkilm: `2.10.0` appears to be the latest version, I'll add that.

apps/docs/tsconfig.json
nkilm: Alright

thibaultleouay:
@nkilm  thanks for the work will merge it 🚀

Issue #142, opened 2023-08-02
Fix migrate script dont run


thibaultleouay:
Should fix the tsconfig.ts to run the migrate script 

As a quick fix run the migration with 

Bun arc/migrate.ts

thibaultleouay:
This has been fixed 🚀

PR #167, opened 2023-08-04
💌 store user email

thibaultleouay:


packages/api/src/router/clerk/webhook.ts
mxkaske, line 1: For already logged in users, we will not be able to access the data with `"user.created"`.

Instead, what we can do is checking if the type is `"session.created"`, taking the `data.user_id` and update our created data by accessing the data that clerk has been saved [via `getUser()`](https://clerk.com/docs/nextjs/user-object#get-user-user-id). This will allow us to fill the gap from time to time.

packages/db/src/schema/user.ts
```diff
@@ -12,6 +12,10 @@ export const user = sqliteTable("user", {
   id: integer("id").primaryKey(),
   tenantId: text("tenant_id", { length: 256 }).unique(), // the clerk User Id
 
+  firstName: text("first_name").default(""),
+  lastName: text("last_name").default(""),
+  email: text("email").default(""),
```
mxkaske, line 17: It would be nice to store the profile image source url as well so we can use it inside of our application.

thibaultleouay:
will also resolve #142

PR #169, opened 2023-08-05
🐛 page slug can't be empty

### LINKED ISSUE #168: Bug: status page is created even when `slug` is not provided⚠
## Demo

[screen-capture.webm](https://github.com/openstatusHQ/openstatus/assets/79012023/e9606ea4-58ab-4604-bc83-359c2c7fac4c)

One solution would be to check if all relevant fields are present in the form before making the form submission. Alterly we can generate a random slug as we are already doing with workspaces.

thibaultleouay:
That will fix #168

### DIFF
--- a/packages/db/src/schema/page.ts
+++ b/packages/db/src/schema/page.ts
@@ -38,23 +38,27 @@ export const pageRelations = relations(page, ({ many, one }) => ({
   }),
 }));
 
+const slugSchema = z
+  .string()
+  .regex(
+    new RegExp("^[A-Za-z0-9-]+$"),
+    "Only use digits (0-9), hyphen (-) or characters (A-Z, a-z).",
+  )
+  .min(3)
+  .toLowerCase();
+
 // Schema for inserting a Page - can be used to validate API requests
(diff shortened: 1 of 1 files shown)


packages/db/src/schema/page.ts
```diff
@@ -54,7 +54,7 @@ export const insertPageSchemaWithMonitors = insertPageSchema.extend({
   customDomain: z.string().optional().default(""),
   monitors: z.array(z.number()).optional(),
   workspaceSlug: z.string().optional(),
-  slug: z.string().toLowerCase(),
+  slug: z.string().toLowerCase().min(1),
```
mxkaske: We [overwrite](https://zod.dev/?id=extend) the `insertPageSchema` and it's defined slug with regex and `min(3)` a few lines above. Let's remove that line entirely and add the `.toLowerCase()` to the `insertPageSchema.slug`.

mxkaske: Or use the same `regex` and `min(3)` as well here.

mxkaske (approved):

Issue #112, opened 2023-07-31
UI design bug on All main pages


kartikk-k:
When we resize window using button from top right between close and minimize, the background stays on the old position after switching to full screen again. 

**Pages:** All pages

<img width="960" alt="image" src="https://github.com/openstatusHQ/openstatus/assets/88406357/215093aa-e0b1-44d7-bc6c-d8eddba87a37">

nkilm:
@kartikk-k Apparently this bug occurs only on chrome and edge browsers not for Firefox.

mxkaske:
Thanks @nkilm for checking and @kartikk-k for reporting. Not having that issue on Safari neither.

Seems like the `svg` is not extending on window resize. Needs a bit more investigation.

nkilm:
@mxkaske It is a bug in the browser itself. I did some investigation online and it's still being fixed. There are some solutions listed which didn't work for me locally. 

References
- https://stackoverflow.com/questions/58088120/image-svg-doesnt-extend-over-the-entire-width-of-the-page
- https://stackoverflow.com/questions/9334095/background-size100-100-doesnt-work-properly-in-chrome
- https://yootheme.com/support/question/6801?order=modified

kartikk-k:
**Problem:** More specifically mask inside the SVG background is creating this bug.
**Solution:** It's a trade-off but conditionally not rendering on Edge and Chrome until it is fixed by browsers. 

@mxkaske will this solution work?

### [Screenshot] DOM element creating bug:
<img width="960" alt="image" src="https://github.com/openstatusHQ/openstatus/assets/88406357/1fe0d28a-cd83-4348-8195-8f7060cb942b">

mxkaske:
Good point. If that's a rendering issue on Edge and Chrome we can hide the `mask` element. It would be still nice to have the same dotted background.

kartikk-k:
Nice, should I fix that and make  a PR?

kartikk-k:
@mxkaske  I also need Figma file with dotted background SVG or dotted background image to fix this issue.

mxkaske:
Sure @kartikk-k, feel free to open a pull request. There is no Figma file or background image I can provide you. 

For the OG-images, we've been using the following style:

```
<div 
    style={{
      backgroundImage: "radial-gradient(#cbd5e1 10%, transparent 10%)",
      backgroundSize: "24px 24px",
    }}
/>
```

Maybe we can use it and update the color to use a css variable.

PR #170, opened 2023-08-06
Fix: background design bug on main pages

kartikk-k:
Issue: #112 

Update main background container's className position to `absolute` from `fixed`

Previous:
```html
<div className="fixed -z-50 h-full w-full">
```

Updated:
```html
<div className="absolute -z-50 h-full w-full">
```


mxkaske (approved):
This is soo good. Thanks @kartikk-k! 🚀

kartikk-k:
Thanks but there has been a minor issue with this change which is fixed in PR #171 with one more related issue.

PR #172, opened 2023-08-08
Fix: eslint import warning on `packages\ui\turbo\generators\config.ts`

ff6k:
Fix eslint type import warning: _All imports in the declaration are only used as types. Use `import type`._

image


thibaultleouay:
Thanks @ff6k  🔥

PR #173, opened 2023-08-09
🐛 add created at

thibaultleouay:


mxkaske:
Can we add a `createdBy` relationship to the user, for all the tables? 😬

thibaultleouay:
> Can we add a `createdBy` relationship to the user, for all the tables? 😬

Why ? 

A user can be added or remove from a workspace ?

mxkaske:
> > Can we add a `createdBy` relationship to the user, for all the tables? 😬
> 
> Why ?
> 
> A user can be added or remove from a workspace ?

Agree, this might not be for security reason but for history track. ‘Who did what’. Can be done at a later stage if we really need it.

PR #175, opened 2023-08-10
 🚀add favicon

thibaultleouay:


apps/web/src/components/forms/status-page-form.tsx
```diff
@@ -175,6 +195,44 @@ export function StatusPageForm({
...
+                      accept="image/x-icon,image/png"
+                      ref={inputFileRef}
+                      onChange={(e) => handleChange(e.target.files)}
+                    />
+                  )}
+                  {field.value && (
+                    <div className="flex items-center">
+                      <div className="h-5 w-5">
+                        <img src={field.value} />
+                      </div>
```
mxkaske: Can we have the `Image` component from Next here? We might want to extend `next.config.js` with the [vercel domain](https://nextjs.org/docs/pages/api-reference/components/image#domains) as well. 

I think a little border would be nice as well

```suggestion
                      <div className="h-5 w-5 border border-border rounded-sm">
                        <img src={field.value} />
                      </div>
```

apps/web/src/app/status-page/[domain]/page.tsx
Comment (mxkaske) [inline apps/web/src/app/status-page/[domain]/page.tsx:1]: We could also display the Icon next to the header for example. Can be done in a later stage.

PR #176, opened 2023-08-10
👨‍⚖️ Some boring legal info

thibaultleouay:


mxkaske (commented):
Can you please add a `layout.tsx` file to it (copied from blog):

```
import * as React from "react";

import { MarketingLayout } from "@/components/layout/marketing-layout";

export default function LegalLayout({
  children,
}: {
  children: React.ReactNode;
}) {
  return <MarketingLayout>{children}</MarketingLayout>;
}
```

We might want to change the Footer a bit as we will add more links and infos to it. Can be done in another stage!

PR #179, opened 2023-08-12
remove tooltip icon when empty

chronark:
Currently the tooltip is hoverable but then nothing shows up if I didn't add a description

image


mxkaske:
Thank you for the fix! 🚀

PR #182, opened 2023-08-14
Improve backend code

thibaultleouay:


packages/api/src/router/incident.ts
```diff
@@ -18,36 +18,29 @@ import {
 } from "@openstatus/db/src/schema";
 
 import { createTRPCRouter, protectedProcedure } from "../trpc";
+import {
+  hasUserAccessToWorkspace as getUserAndWorkspace,
+  hasUserAccessToWorkspace,
```
mxkaske: Why having both?

packages/api/src/router/incident.ts
```diff
@@ -18,36 +18,29 @@ import {
...
-        .select()
-        .from(usersToWorkspaces)
-        .where(eq(usersToWorkspaces.workspaceId, currentWorkspace.id))
-        .innerJoin(currentUser, eq(usersToWorkspaces.userId, currentUser.id))
-        .get();
-
+      const result = await hasUserAccessToWorkspace({
+        workspaceSlug: opts.input.workspaceSlug,
+        tenantId: opts.ctx.auth.userId,
+        ctx: opts.ctx,
```
mxkaske: Can we **only** pass the prop `opts`? At least the `ctx` as auth is always included. No need to pass `tenantId` as extra property.

packages/api/src/router/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @returns {Promise<{ workspaceId: string; userId: string }> | undefined} - workspaceId and userId
+ */
+export const hasUserAccessToWorkspace = async ({
+  workspaceSlug,
+  tenantId,
+  ctx,
+}: {
+  workspaceSlug: string;
+  tenantId: string;
+  ctx: Context;
```
mxkaske, line 18: Here e.g. we can access tenantId by `ctx.auth.tenantId`, no need to pass it extra.

packages/api/src/router/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+    .where(eq(workspace.slug, workspaceSlug))
+    .get();
+  if (!currentWorkspace) return;
+  const result = await ctx.db
+    .select()
+    .from(usersToWorkspaces)
+    .where(eq(usersToWorkspaces.workspaceId, currentWorkspace.id))
+    .innerJoin(currentUser, eq(usersToWorkspaces.userId, currentUser.id))
+    .get();
+  // the user don't have access to this workspace
```
mxkaske:
```suggestion
  // the user doesn't have access to this workspace
```

mxkaske (commented):
Good stuff 🔥

packages/api/src/router/incident.ts
thibaultleouay: Issue will be fixed 🤣

mxkaske (commented):
lfg! 🚀

PR #186, opened 2023-08-15
feat: clean up

mxkaske:
part 1

### DIFF
--- a/apps/web/src/components/layout/app-footer.tsx
+++ b/apps/web/src/components/layout/app-footer.tsx
@@ -1,0 +1,30 @@
+import Link from "next/link";
+
+import { Shell } from "../dashboard/shell";
+
+export function AppFooter() {
+  return (
+    <footer className="w-full">
+      <Shell className="flex items-center justify-between">
+        <div className="text-muted-foreground text-xs font-light">
+          All rights reserved &copy;
+        </div>
+        <div className="text-right text-xs">
+          <Link
+            href="/legal/terms"
+            className="text-foreground underline underline-offset-4 hover:no-underline"
+          >
+            Terms
+          </Link>
+          <span className="text-muted-foreground/70 mx-1">&bull;</span>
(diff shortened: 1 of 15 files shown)


apps/web/src/components/layout/app-footer.tsx
```diff
@@ -0,0 +1,30 @@
...
+          <Link
+            href="/legal/terms"
+            className="text-foreground underline underline-offset-4 hover:no-underline"
+          >
+            Terms
+          </Link>
+          <span className="text-muted-foreground/70 mx-1">&bull;</span>
+          <Link
+            href="/legal/privacy"
+            className="text-foreground underline underline-offset-4 hover:no-underline"
```
thibaultleouay, line 22: I think legacy and tos don't need to be on every page when you are logged in

apps/web/src/components/marketing/plans.tsx
```diff
@@ -1,96 +1,44 @@
...
   );
 }
 
-interface Props extends Plan {
+interface Props extends PlanProps {
   className?: string;
 }
 
-function Plan({
+export function Plan({
```
thibaultleouay, line 33: We are using it also in setting maybe extract it to components shared plan

thibaultleouay (approved):

apps/web/src/components/layout/app-footer.tsx
mxkaske: Good point, a lot of dashboard don't even have a footer. Maybe we can move it only to marketing pages

thibaultleouay: Yes 🔥

apps/web/src/components/marketing/plans.tsx
mxkaske: We now use only one component for both, marketing and settings. Will extract it later!

PR #197, opened 2023-08-17
fix: limits

mxkaske:


packages/api/src/router/utils.ts
```diff
@@ -39,5 +40,11 @@ export const hasUserAccessToWorkspace = async ({
...
   if (!result || !result.users_to_workspaces) return;
 
-  return { workspace: currentWorkspace, user: result.currentUser };
+  const plan = (currentWorkspace.plan || "free") as "free" | "pro"; // FIXME: that is a hotfix
+
+  return {
+    workspace: currentWorkspace,
+    user: result.currentUser,
+    plan: allPlans[plan],
+  };
```
mxkaske, line 49: Extending what we return here to make it reusable! 🔥

packages/api/src/router/monitor.ts
```diff
@@ -163,6 +168,16 @@ export const monitorRouter = createTRPCRouter({
...
 
+      const currentWorkspace = await opts.ctx.db.query.workspace.findFirst({
+        where: eq(workspace.id, result.users_to_workspaces.workspaceId),
+      });
+
+      if (!currentWorkspace) return;
+
+      const plan = (currentWorkspace?.plan || "free") as "free" | "pro";
+
+      const periodicityLimit = allPlans[plan].limits.periodicity;
```
mxkaske, line 179: Not ideal, but the whole query needs to be refactored!

apps/web/src/app/app/(dashboard)/[workspaceSlug]/status-pages/page.tsx
```diff
@@ -33,7 +33,7 @@ export default async function Page({
   });
 
   const isLimit =
-    (monitors?.length || 0) >=
+    (pages?.length || 0) >=
     allPlans[workspace?.plan || "free"].limits["status-pages"];
```
Comment (mxkaske) [inline apps/web/src/app/app/(dashboard)/[workspaceSlug]/status-pages/page.tsx:37]: We should move the logic of getting the limits into `getWorkspaceBySlug`.

PR #204, opened 2023-08-18
add migrate script to make database migrations before starting dev server

### LINKED ISSUE #203: Run migrations before starting the development server
Currently, the documentation does not state that one needs to run migrations before starting the server but it is a requirement.

AmoabaKelvin:
Fixes #203

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "private": true,
   "scripts": {
     "build": "turbo run build",
-    "dev": "turbo run dev",
+    "migrate": "cd packages/db && pnpm migrate",
+    "dev": "pnpm migrate && turbo run dev",
     "lint": "turbo run lint",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,7 +2,8 @@
   "private": true,
   "scripts": {
     "build": "turbo run build",
-    "dev": "turbo run dev",
+    "migrate": "cd packages/db && pnpm migrate",
+    "dev": "pnpm migrate && turbo run dev",
```
mxkaske, line 6: I am not sure if we should run it inside of the root `package.json`. It feels to be more appropriate to run inside of `/apps/web`, wdyt?

AmoabaKelvin: You're spot on with that. I think I was thinking in another direction but things are much clearer now. 

I guess we can go ahead and close this.

mxkaske: I'll close this one for now!

PR #205, opened 2023-08-18
appending open-status header to avoid tracking requests if this header exist

### LINKED ISSUE #114: provide a way to un track these IP from analytics
this way we lead to too much miss good analytics, as edge will use different ips each time even for the same region

so is there's a way to prevent that ?!

### LINKED ISSUE #190: [OPE-77] Provide a way to untrack ping requests (e.g. from analytics)
We could think of adding an additional request header property `"openstatus-no-track"` or a custom user agent.

Seen in Steven's [middleware](https://github.com/steven-tey/dub/blob/04c5401fcd993c64134ddfc91df775af7b578d12/lib/middleware/link.ts#L32).

Feature Request #114

<sub>From [SyncLinear.com](https://synclinear.com) | [OPE-77](https://linear.app/openstatus/issue/OPE-77/provide-a-way-to-untrack-ping-requests-eg-from-analytics)</sub>

Mahmoudgalalz:


Mahmoudgalalz:
@mxkaske can you take a look, and if there's smth else ask me to do it to complete this PR
#190
should we add this option to the user dashboard to enable or disable it?

apps/web/src/app/api/ping/route.ts
```diff
@@ -2,6 +2,7 @@ import { NextResponse } from "next/server";
 
 export const dynamic = "force-dynamic";
 
-export async function GET() {
+export async function GET(req:Request) {
+  req.headers.set('OpenStatus-Ping',true)
```
mxkaske: Hey @Mahmoudgalalz, this is the endpoint that we ping and want to know if it contains a `req.headers.get('OpenStatus-Ping') for example.

In here, we are calling all the endpoints to ping (e.g. `/api/ping`) and that's where we will need to add the header.
https://github.com/openstatusHQ/openstatus/blob/4242d8bd92c973e122b73a10a003464afb300adf/apps/web/src/app/api/checker/regions/_checker.ts#L82

mxkaske:
> should we add this option to the user dashboard to enable or disable it?

I don't think we need it. It's 'just' a header that will be appended for every request. Up to the user to use it or not to prevent conditionally the tracking events.

apps/web/src/app/api/ping/route.ts
Mahmoudgalalz: I have been looking inside the _checker all the time, but yeah I will do it rn and remove

so now there's no point form checking the header in the ping right ?!

Mahmoudgalalz:
@mxkaske check this

apps/web/src/app/api/ping/route.ts
```diff
@@ -2,6 +2,6 @@ import { NextResponse } from "next/server";
 
 export const dynamic = "force-dynamic";
 
-export async function GET() {
+export async function GET(req:Request) {
```
thibaultleouay: I think the req:Request could be removed

mxkaske:
I have resolved the merge conflict due to the addition of `headers` in #210.

apps/web/src/app/api/checker/regions/_checker.ts
```diff
@@ -87,7 +87,10 @@ export const checker = async (request: Request, region: string) => {
     const res = await fetch(result.data?.url, {
       method: result.data?.method,
       cache: "no-store",
-      headers: { ...headers },
+      headers: { 
+         "OpenStatus-Ping":"true",
+        ...headers 
+      },
```
mxkaske:
```suggestion
      headers: { 
        "OpenStatus-Ping": "true",
        ...headers 
      },
```

apps/web/src/app/api/ping/route.ts
```diff
@@ -2,7 +2,7 @@ import { NextResponse } from "next/server";
...
```
mxkaske:
```suggestion
export async function GET() {
```

mxkaske:
Thanks @Mahmoudgalalz! 🔥

PR #227, opened 2023-08-23
feat: added dependabot, husky and lint-staged

adithyaakrishna:
Description: 

- Added `depenabot.yml` for auto package updates
- Added husky and lint-staged for pre-commit hooks
- Added engines to enforce Node v18.0.0 and above


mxkaske:
Appreciate it! Thanks @adithyaakrishna. We will check it later the day. 🙌

mxkaske:
Sorry @adithyaakrishna, I missed that one. 
Thank you for providing this PR! 🚀

Personally, I am not a big fan of adding dependabot in a project under development as we will be spammed with new updates. I bet we will still need it. The addition of `husky` and `lint-staged` is great.  🔥

Thankfully you haven't developed conventional pre-commit commit-msg for @thibaultleouay. 🤣

thibaultleouay:
I'm not a huge fan of pre-commit hooks they are just slowing me down one my local dev, we have some github action doing the check IMO 

Also when we had new dependancies exact should be the default :) 


What's the advantage of dependabot over renovate ( I have used renovate in an other project )

mxkaske:
Hey @adithyaakrishna, appreciate it a lot! We will merge it once needed! 💪

PR #228, opened 2023-08-23
feat: regex custom domain and input with addons

mxkaske:
Regex to ensure that the input does **not** start with `https://`, `http://` or `www`. Update includes a new `InputWithAddons` component for `trailing` and `leading` text/icon/...

CleanShot 2023-08-23 at 22 03 38@2x

CleanShot 2023-08-23 at 22 03 17@2x

### DIFF
--- a/packages/db/src/schema/page.ts
+++ b/packages/db/src/schema/page.ts
@@ -50,15 +50,22 @@ const slugSchema = z
   .min(3)
   .toLowerCase();
 
+const customDomainSchema = z
+  .string()
+  .regex(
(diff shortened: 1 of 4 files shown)


packages/db/src/schema/page.ts
```diff
@@ -50,15 +50,22 @@ const slugSchema = z
   .min(3)
   .toLowerCase();
 
+const customDomainSchema = z
+  .string()
+  .regex(
+    new RegExp("^(?!https?://|www.)([a-zA-Z0-9]+(.[a-zA-Z0-9]+)+.*)$"),
+    "Should not start with http://, https:// or www.",
```
thibaultleouay, line 57: why not www ?

PR #236, opened 2023-08-26
fix: only regenerate slug if there is no default

### LINKED ISSUE #231: bug: auto-generate slug
With #198 we have introduced a bug and temporarily disabled the functionality for now.

The reason: If your slug was different to what would have been generated, whenever you would re-edit a page, it would `slugify` the title on every form opening.

How to reproduce:

1. uncomment the lines here:
https://github.com/openstatusHQ/openstatus/blob/7a80fbdd643174f20e2d90c3bc3d63422652b646/apps/web/src/components/forms/status-page-form.tsx#L95
2. create a status page (edited) with a `title` and change the `slug` that has been auto generated - and save
3. edit the same status page (edited) - the `slug` should have been updated to the auto generated one, not the one saved

AmoabaKelvin:
Fixes #231

### DIFF
--- a/apps/web/src/components/forms/status-page-form.tsx
+++ b/apps/web/src/components/forms/status-page-form.tsx
@@ -94,10 +94,10 @@ export function StatusPageForm({
   }, [checkUniqueSlug]);
 
   useEffect(() => {
-    if (!watchSlug) {
+    if (!defaultValues?.title) {
       form.setValue("slug", slugify(watchTitle));
     }
-  }, [watchTitle, form, watchSlug]);
+  }, [watchTitle, form, watchSlug, defaultValues?.title]);
 
   const onSubmit = async ({
     ...props
--- a/apps/web/src/components/forms/status-page-form.tsx
+++ b/apps/web/src/components/forms/status-page-form.tsx
@@ -97,7 +97,7 @@ export function StatusPageForm({
     if (!defaultValues?.title) {
       form.setValue("slug", slugify(watchTitle));
     }
-  }, [watchTitle, form, watchSlug, defaultValues?.title]);
+  }, [watchTitle, form, defaultValues?.title]);
 
   const onSubmit = async ({
     ...props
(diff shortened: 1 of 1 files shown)


apps/web/src/components/forms/status-page-form.tsx
```diff
@@ -92,12 +93,11 @@ export function StatusPageForm({
...
-  // useEffect(() => {
-  //   if (!watchSlug) {
-  //     form.setValue("slug", slugify(watchTitle));
-  //   }
-  // }, [watchTitle, form, watchSlug]);
+  useEffect(() => {
+    if (!watchSlug) {
+      form.setValue("slug", slugify(watchTitle));
+    }
+  }, [watchTitle, form, watchSlug]);
```
mxkaske: Thanks for tackling that issue! 🚀

This will trigger the `.setValue` function only if the `slug` is undefined or an empty string in our case. If the user deletes the entire slugified content, the function will get fired and automatically resets the `slug` from `undefined` to the slugified content.

I think we should track if the `defaultValue.slug` exists. Only if not, we watch the changes from the `title` input to update the `slug` input. Changing the `slug` afterwards is possible as the `useEffect` will not get fired. Only if, on creation (as `defaultValues` is undefined)m the `title` is changed again after having updated the auto-generated `slug`, it will re-change the already updated `slug`. (Getting inception vibes here.)

Let me know if that works!

```suggestion
  useEffect(() => {
    if (!defaultValues?.title) {
      form.setValue("slug", slugify(watchTitle));
    }
  }, [watchTitle, form, defaultValues]);
```

apps/web/src/components/forms/status-page-form.tsx
```diff
@@ -92,12 +93,11 @@ export function StatusPageForm({
...
-  // useEffect(() => {
-  //   if (!watchSlug) {
-  //     form.setValue("slug", slugify(watchTitle));
-  //   }
-  // }, [watchTitle, form, watchSlug]);
+  useEffect(() => {
+    if (!defaultValues?.title) {
+      form.setValue("slug", slugify(watchTitle));
+    }
+  }, [watchTitle, form, watchSlug, defaultValues?.title]);
```
mxkaske: You'll need to remove `watchSlug` from deps, otherwise we won't be able to customize the slug on page creation. Otherwise, lgtm!

```suggestion
  }, [watchTitle, form, defaultValues?.title]);
```

mxkaske:
Thanks Kelvin! 🔥

PR #238, opened 2023-08-26
chore: better toast, tracker fix and more

mxkaske:
lots of small toast improvements incl. error message with discord support button and extracting in separate component
CleanShot 2023-08-26 at 20 54 45@2x

fixing ui bug
CleanShot 2023-08-27 at 00 00 25@2x

fixing empty headers on advanced monitor settings: until now, we would have saved and send the empty header to the endpoint. now checking if the `"key"` is not an empty string.

replace placeholder to "Documenso"

removing redirect to overview page after save
> This needs to discuss. But as we are getting more and more tabs, we should update the way we are navigating! @thibaultleouay


thibaultleouay (commented):
LGTM 🔥

PR #239, opened 2023-08-26
fix: Only show body input when method is POST

### LINKED ISSUE #232: [OPE-86] Only show `body` Input when advanced monitor settings method is "POST"
The `body` textarea should be `hidden` if "GET" is chosen. Otherwise, if a user is selecting the "GET" method and adds a body, we will get:

```
[TypeError: Request with GET/HEAD method cannot have body.]
```

<sub>From [SyncLinear.com](https://synclinear.com) | [OPE-86](https://linear.app/openstatus/issue/OPE-86/only-show-body-input-when-advanced-monitor-settings-method-is-post)</sub>

zahid47:
fixes #232

### DIFF
--- a/apps/web/src/components/forms/advanced-monitor-form.tsx
+++ b/apps/web/src/components/forms/advanced-monitor-form.tsx
@@ -71,7 +71,7 @@ export function AdvancedMonitorForm({ defaultValues, workspaceSlug }: Props) {
       method: defaultValues?.method ?? "GET",
     },
   });
-  const methodWatch = form.watch("method");
+  const watchMethod = form.watch("method");
   const router = useRouter();
   const searchParams = useSearchParams();
   const monitorId = searchParams.get("id");
@@ -209,7 +209,7 @@ export function AdvancedMonitorForm({ defaultValues, workspaceSlug }: Props) {
             </FormItem>
           )}
         />
-        {methodWatch === "POST" && (
+        {watchMethod === "POST" && (
           <div className="sm:col-span-4 sm:col-start-1">
             <FormField
               control={form.control}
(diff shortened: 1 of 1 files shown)


apps/web/src/components/forms/advanced-monitor-form.tsx
```diff
@@ -178,9 +185,10 @@ export function AdvancedMonitorForm({ defaultValues, workspaceSlug }: Props) {
             <FormItem className="sm:col-span-2 sm:col-start-1 sm:self-baseline">
               <FormLabel>Method</FormLabel>
               <Select
-                onValueChange={(value) =>
-                  field.onChange(methodsEnum.parse(value))
-                }
+                onValueChange={(value) => {
+                  field.onChange(methodsEnum.parse(value));
+                  form.resetField("body");
```
mxkaske: We will need an additional config here, otherwise it will be back to `defaultValue.body`, which might be defined if we switch from "POST" back to "GET". (see [docs](https://react-hook-form.com/docs/useform/resetfield)).
```suggestion
                  form.resetField("body", { defaultValue: "" });
```

mxkaske (commented):
besides the one comment, lgtm!

apps/web/src/components/forms/advanced-monitor-form.tsx
zahid47: Yep, I totally missed that case. Will push another commit.

apps/web/src/components/forms/advanced-monitor-form.tsx
```diff
@@ -65,6 +71,7 @@ export function AdvancedMonitorForm({ defaultValues, workspaceSlug }: Props) {
       method: defaultValues?.method ?? "GET",
     },
   });
+  const methodWatch = form.watch("method");
```
mxkaske: Good thing on the rename! I'd suggest to go for `watchMethod` as this is our current naming convention.

mxkaske:
Thanks @zahid47! 🚀

mxkaske (approved):

PR #244, opened 2023-08-27
Move user to monitor edit from status page.

AmoabaKelvin:
When a user clicks on the monitor from the status page, let's redirect them to the edit page of that particular monitor so they can transition between pages swiftly.

So once `my website` in the image below is clicked, we move them to the monitor edit page of that particular monitor.
image


AmoabaKelvin:
Forgot to checkout to main before switching to a new branch, so the previous grammar fix was included in the commits

mxkaske:
Love it! Yeah, no worries regarding grammar stuff.

PR #261, opened 2023-08-28
🧹 remove sentry dsn

thibaultleouay:


apps/web/.env.example
```diff
@@ -49,4 +49,6 @@ PROJECT_ID_VERCEL=
 TEAM_ID_VERCEL=
 VERCEL_AUTH_BEARER_TOKEN=
 
-NEXT_PUBLIC_URL=
\ No newline at end of file
+NEXT_PUBLIC_URL=
+
+NEXT_PUBLIC_SENTRY_DSN=''
```
mxkaske, line 54: Let's add `SENTRY_AUTH_TOKEN` as well (including `env.ts`)

thibaultleouay: that's only to publish the source map to sentry not sure if we really need it

PR #264, opened 2023-08-28
Feat/features homepage

mxkaske:
Just leaving it here - if we ever need it.

CleanShot 2023-08-28 at 23 40 56

Also includes a `loading` state for custom domain fetching.


mxkaske:
Will not be used.

Issue #263, opened 2023-08-28
[OPE-92] `/oss-friends` page


mxkaske:
Let's be one of the cool kids and have a simple list with all of them.

We could have a `/config/oss-friends.ts`configuration with the following item list props:

```
type Friend = {
  name: string;
  website: string;
  description: string;
  socialLinks: Record<"twitter" | "github" | "discord" | "...", string>;
  logo: string;
  // and more if there is
}
```

Use the `MarketingLayout` as `layout.tsx` (similar to blogs) and create a grid with what number the `grid-cols-*` looks good. We might want to reduce the number on smaller devices.

Lets try to use shadcn/ui a maximum and use a `Card` component (for every `Friend`) as they have `title` and `description`. The content can include icons for the `socialLinks` and a primary button for the `website`.

[https://twitter.com/formbricks/status/1660735970281508878?s=20](https://twitter.com/formbricks/status/1660735970281508878?s=20)

At the end, lets add the "OSS Friends" link in our footer under "Community"

<sub>From [SyncLinear.com](https://synclinear.com) | [OPE-92](https://linear.app/openstatus/issue/OPE-92/oss-friends-page)</sub>

PR #269, opened 2023-08-30
Open source friends page

AmoabaKelvin:
Completes #263


apps/web/src/config/ossFriends.ts
mxkaske, line 1: To stick to our file naming convention, let's go for `oss-friends.ts`

apps/web/src/app/oss-friends/page.tsx
```diff
@@ -0,0 +1,72 @@
...
+                </CardTitle>
+              </div>
+              <CardDescription>{friend.description}</CardDescription>
+            </CardHeader>
+            <CardFooter>
+              <div className="flex items-center gap-2.5">
+                <Link href={friend.website} target="_blank">
+                  <Icons.globe className="text-muted-foreground h-6 w-6 hover:text-black" />
+                </Link>
+                {friend.socialLinks.twitter && (
```
mxkaske: I wonder if we can do

```
Object.entries(friends.socialLinks).map(([key, value]) => <Link key={key} {...} />)
```

To future proof more links

mxkaske (commented):
Small remarks. Great work! 🔥

apps/web/src/app/oss-friends/page.tsx
AmoabaKelvin: Updated them all @mxkaske

apps/web/src/config/oss-friends.ts
mxkaske, line 1: We can type safe it a bit more by moving `TwitterIcon` to the `Icons` and use the `ValidIcon` property.

```ts
import { Icons, ValidIcon } from "@/components/icons";

type SocialLink = {
  icon: ValidIcon;
  link: string;
};
```

You will only need to update the `icon` to use the object key
```
twitter: {
  icon: "twitter",
  link: "https://twitter.com/documenso",
},
```

and inside of the `CardFooter` use

```ts
const Icon = Icons[value.icon];
```

mxkaske (commented):
Last comment to remove `any` and maybe we can reduce the size of the icons a bit. Other than that lgtm!

PR #272, opened 2023-09-02
feat: vercel log drains

mxkaske:
How to use "Log Drains" manually

1. Verify your endpoint

- Define the sources and add the webhook endpoint: `URL/api/integrations/vercel`.
- Add the `INTEGRATION_SECRET` to your web `.env`
- Comment the `if (bodySignature !== request.headers.get("x-vercel-signature"))` block (we will need it in step two)
- Click on "Test Log Drain" and check if the request successfully works
- Click on "Verify"

> No need to add custom headers for now. This is secure enough! Maybe in the future.

CleanShot 2023-09-02 at 14 12 59@2x

2. Add your signature

- Add the secret to `LOG_DRAIN_SECRET` in the web `.env`
- Uncomment the `if (bodySignature !== request.headers.get("x-vercel-signature"))` block

CleanShot 2023-09-02 at 14 15 10@2x

Et voila, you should get the vercel log drain data. It is parsed with `zod` and can be published to tb by uncommenting

```js
// logDrains.data.map((log) => publishVercelLogDrain()(log));
```

CleanShot 2023-09-02 at 12 54 11@2x


thibaultleouay:
@mxkaske  I have updated the pr, if you can check it ! 

I will add the secret and create a prod integration on Vercel !

apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/page.tsx
```diff
@@ -0,0 +1,44 @@
...
+            href={
+              workspace?.id === 1
+                ? "https://vercel.com/integrations/openstatus-staging/new"
+                : "#"
+            }
+            target="_blank"
+            key={"vercel"}
+          >
+            <Button>{workspace?.id === 1 ? "Configure" : "Coming soon"}</Button>
+          </a>,
```
Comment (mxkaske) [inline apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/page.tsx:39]: To have a link, acting like it but looking like a button, you can
```jsx
<Button asChild>
  <a href={`${url}/app`}>Configure</a>
</Button>
```

apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/vercel/configure/page.tsx
Comment (mxkaske) [inline apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/vercel/configure/page.tsx:1]: Can we move most of it into the `@openstatus/vercel` package, as it is mostly the `Configure` page from the package.

Also, why would we need that page? We already have the `/configure` page from the vercel integration. This seems to be a duplicate and imo, is an overkill for now and can be removed.

apps/web/src/middleware.ts
```diff
@@ -109,10 +113,42 @@ export default authMiddleware({
...
         // return NextResponse.redirect(new URL("/app/onboarding", req.url));
         // probably redirect to onboarding
         // or find a way to wait for the webhook
       }
+      console.log("redirecting to onboarding");
+      return;
+    }
+    if (
+      auth.userId &&
+      req.nextUrl.pathname === "/app/integrations/vercel/configure"
```
mxkaske, line 129: Why do we need this if statement?

apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/vercel/configure/page.tsx
```diff
@@ -0,0 +1,122 @@
...
+    console.log(project.id);
+    api.integration.createIntegration.mutate({
+      workspaceSlug: params.workspaceSlug,
+      input: {
+        name: "Vercel",
+        data: JSON.stringify(project),
+        externalId: project.id,
+      },
+    });
+  }
```
Comment (mxkaske) [inline apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/vercel/configure/page.tsx:51]: This will create multiple integrations for the same project on each page load. - Ok just saw we are doing the check in `createIntegration`. I'd prefer not to call the mutation instead.

packages/integrations/vercel/src/routes/callback.ts
```diff
@@ -0,0 +1,54 @@
...
+  const token = await getToken(code);
+
+  // TODO: automatically install log drains to the allowed projects
+
+  const iv = crypto.randomBytes(16);
+  const encryptedToken = encrypt(iv, Buffer.from(token));
+
+  // redirect to vercel's integration page after installation
+  const res = NextResponse.redirect(
+    `https://wwww.openstatus.dev/app/integrations/vercel/configure`,
```
mxkaske, line 28: Why not using the `next` URL to redirect the user to Vercel's integration page?

packages/tinybird/datasources/vercel_log_drain.datasource
```diff
@@ -0,0 +1,29 @@
...
+`proxy_timestamp` Int64 `json:$.proxy.timestamp` ,
+`proxy_status` String `json:$.proxy.status` ,
+`proxy_userAgent` Array(String) `json:$.proxy.userAgent[:]` ,
+`requestId` String `json:$.requestId` ,
+`type` String `json:$.type` ,
+`source` String `json:$.source` ,
+`timestamp` Int64 `json:$.timestamp` ,
+
+ENGINE "MergeTree"
+ENGINE_SORTING_KEY "requestId, projectId"
```
mxkaske, line 28: Let's check again what the important sorting keys are. Maybe `type` and `source` are relevant as well? `requestId`  will not be needed I think.

mxkaske (commented):
I think we can simplify the integration for the beginning by removing the `/integrations/vercel` pages inside of our app as the integration will be configured inside of vercel itself.

apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/vercel/configure/page.tsx
Comment (thibaultleouay) [reply apps/web/src/app/app/(dashboard)/[workspaceSlug]/integrations/vercel/configure/page.tsx:1]: In this page 
It's the page we are redirected from vercel 

And we connect a Vercel project to a workspace

apps/web/src/middleware.ts
thibaultleouay: In the redirect url server side from Vercel webhook we don't know the workspace associated nor the user 

We redirect to app integration then on the middleware we redirect to the configuration specific page

packages/integrations/vercel/src/routes/callback.ts
thibaultleouay: Because we need to link a Vercel project with a workspace 😭

packages/tinybird/datasources/vercel_log_drain.datasource
thibaultleouay: Yep its a bit random 

Love to really update it

PR #279, opened 2023-09-06
Show tooltip for disabled button

AmoabaKelvin:
When a button is disabled, show additional tooltip content to the the user indicating why the button has been disabled.

### DIFF
--- a/apps/web/src/components/ui/button-with-disable-tooltip.tsx
+++ b/apps/web/src/components/ui/button-with-disable-tooltip.tsx
@@ -1,3 +1,4 @@
+import { cn } from "@/lib/utils";
 import { Button } from "./button";
 import type { ButtonProps } from "./button";
 import {
@@ -30,7 +31,9 @@ export const ButtonWithDisableTooltip = ({
     return (
       <TooltipProvider>
         <Tooltip>
-          <TooltipTrigger>{ButtonComponent}</TooltipTrigger>
+          <TooltipTrigger asChild>
+            <span tabIndex={0}>{ButtonComponent}</span>
+          </TooltipTrigger>
           <TooltipContent>{tooltip}</TooltipContent>
         </Tooltip>
       </TooltipProvider>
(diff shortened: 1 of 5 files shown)


apps/web/src/components/ui/button.tsx
mxkaske, line 1: Hey @AmoabaKelvin! 

Thanks! Let's keep the shadcn ui components as 'neutral' as possible in case of updates. 

Let's create a new UI component for that: `button-with-disable-tooltip.tsx` (to make it explicit) and basically just extend the button props.

```tsx
interface ButtonWithDisableTooltipProps extends ButtonProps {
  tooltip: string;
}
```

mxkaske (commented):
We can use the same component for the `status-page` as well.

apps/web/src/components/ui/button-with-disable-tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+    // we don't want to show the tooltip if the button is enabled.
+    return (
+      <TooltipProvider>
+        <Tooltip>
+          <TooltipTrigger>{ButtonComponent}</TooltipTrigger>
+          <TooltipContent>{tooltip}</TooltipContent>
+        </Tooltip>
+      </TooltipProvider>
+    );
+  }
```
mxkaske: We are missing the `asChild`, the `span` and the `"pointer-events-none"` classname (maybe use `cn` to append it if disabled) for the disabled component (see [radix-ui](https://www.radix-ui.com/primitives/docs/components/tooltip#displaying-a-tooltip-from-a-disabled-button)). Once done, we can ship it!

AmoabaKelvin: Worked on that, please check it up and let me know what you think

apps/web/src/components/ui/button-with-disable-tooltip.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const ButtonWithDisableTooltip = ({
+  tooltip,
+  disabled,
+  ...props
+}: ButtonWithDisableTooltipProps) => {
+  const ButtonComponent = (
+    <Button
+      {...props}
+      disabled={disabled}
+      className={cn(disabled && "pointer-events-none")}
```
mxkaske: We'll need to deconstruct the props a bit more.

```suggestion
      className={cn(className, disabled && "pointer-events-none")}
```

PR #283, opened 2023-09-07
redirect to monitor creation form from status page

AmoabaKelvin:
When the user has no active monitors and is on the status page, rather than redirecting to the monitors page, let's move the user straight to the monitor creation form, skipping one step!


mxkaske (commented):
such simple UX improvement! Amazing. 🚀

PR #292, opened 2023-09-09
moved components from web  into ui

abdullah-25:


mxkaske:
Hey @abdullah-25, we have merged #309 that includes a big refactoring of the ui components into the `@openstatus/ui` package.

PR #309, opened 2023-09-17
😱 extract ui 💅

thibaultleouay:
extract shadcn ui to it's own package then i can re use it in the vercel integration


packages/ui/src/lib/utils.ts
mxkaske, line 1: We don't need all the functions here. Only use the active once.

mxkaske (commented):
Is it possible to import from `@openstatus/ui` and remove the trailing `src/component`?

PR #318, opened 2023-09-20
Initial localization of status pages to es, fr, de

zackrw:


zackrw:
Abandoning in favor of https://github.com/openstatusHQ/openstatus/pull/451

PR #327, opened 2023-09-24
status filter ui component update

### LINKED ISSUE #221: [OPE-84] Replace filter status code input field with combox like 'region'
There is no need for the status code input. We should replace it with the similar component/behavior like the region one.

The only difficulty is do create the select list dynamically, based on the content we get.

It can be that at some point, we have status `200`, `500`, `429` and another one where we have `200`, `500`, `404` - the select options should reflect accordingly.

CleanShot 2023-09-24 at 18 41 00@2x

<sub>From [SyncLinear.com](https://synclinear.com) | [OPE-84](https://linear.app/openstatus/issue/OPE-84/replace-filter-status-code-input-field-with-combox-like-region)</sub>

hrutik7:
right now I just updated ui and plan to just create a draft as soon as Tinybird issue resolved It will be ready for review


hrutik7:
@mxkaske The ui is already set .
but I am not getting any kind of response in the table 
https://jam.dev/c/ada66c38-a4d2-42ab-91a9-8a63362f2183

apps/web/src/data/code-dictionary.ts
```diff
@@ -0,0 +1,26 @@
+export const codesDict = {
+    
+    
+    arn1: {
```
mxkaske: It would be good to have the key matching the `code` (e.g. here 200) instead of a random edge function.

apps/web/src/components/data-table/data-table-filter-dropdown.tsx
mxkaske, line 1: Can we use the `DataTableFacetedFilter` component as it is instead of creating a new very similar component.

mxkaske (commented):
Thanks for your PR, @hrutik7! Please run `pnpm format` to make sure to be correctly formatted. Feel free to let me know if you have any questions! 😃

apps/web/src/components/data-table/data-table-filter-dropdown.tsx
hrutik7: sure

apps/web/src/data/code-dictionary.ts
hrutik7: sure

apps/web/src/components/data-table/data-table-filter-dropdown.tsx
mxkaske, line 1: This file is not used anymore. Can be removed.

apps/web/src/data/code-dictionary.ts
```diff
@@ -0,0 +1,22 @@
+export const codesDict = {
+  "200": {
+    code: "200",
+    name: "OK",
+    status_code: "200",
```
mxkaske: What is the difference between `code` and `status_code`?

As code convention, we use `camelCase` instead of `snake_case` (e.g. `statusCode`).

apps/web/src/data/code-dictionary.ts
mxkaske, line 1: It would be great to include more common status codes like `400`, `401` e.g. but not every status is relevant.

apps/web/src/components/data-table/data-table-toolbar.tsx
```diff
@@ -22,7 +21,19 @@ export function DataTableToolbar<TData>({
...
+          <DataTableFacetedFilter
+            column={table.getColumn("statusCode")}
+            title="Filter by Status Code"
+            options={Object.keys(codesDict).map((key) => {
+              const typedKey = key as keyof typeof codesDict;
+              return {
+                label: codesDict[typedKey].code,
+                value: codesDict[typedKey].status_code,
+              };
+            })}
```
mxkaske: Is it possible to only display the `codesDict` keys that exist in our data. Supposedly, we only have `200`, we shouldn't display any other option.

It should be possible to access the data via the `table` attribute and reduce the status code.

mxkaske (commented):
Thanks @hrutik7! Left some remarks. Let me know if you have any questions, especially regarding accessing the data via the `table` prop.

apps/web/src/components/data-table/data-table-filter-dropdown.tsx
hrutik7: oh sorry missed this

apps/web/src/data/code-dictionary.ts
hrutik7: sure

apps/web/src/components/data-table/data-table-toolbar.tsx
hrutik7: so the status data will come from the url right ?

mxkaske: The data we display is coming from the `getResponseList()`. One row has all the props from `tbBuildResponseList` including `statusCode`.

It would be nice to do something like:

```ts
const existingStatusCodes = data.reduce((prev, curr) => {
  if (!prev.includes(curr.statusCode)) {
    return [...prev, curr.statusCode]
  }
  return prev
}, [])
```

And then only display the existing status codes instead of all of them inside of the `DataTableFacetedFilter`.

hrutik7: ohhh ok gotcha 
thanks for explaining !

PR #336, opened 2023-09-27
Show confirmation modal when saving a monitor with a  failed ping

AmoabaKelvin:
Sometimes, a user might enter an invalid link when inserting a monitor and they might not necessarily hit the `test Request` button. We now check the ping status of the link entered by the user, and if it was not successful, we show the user the modal to indicate whether they still want to inset the link


apps/web/src/components/modals/failed-ping-alert-confirmation.tsx
```diff
@@ -0,0 +1,63 @@
...
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "@openstatus/ui";
+
+import { LoadingAnimation } from "@/components/loading-animation";
+import type { MonitorProps } from "../forms/montitor-form";
+
+type FailedPingAlertConfirmationProps = {
+  submit: ({ ...props }: MonitorProps) => void;
+  props: MonitorProps;
```
mxkaske: We should call it `monitor`. Calling it `props` can be misleading.

apps/web/src/components/modals/failed-ping-alert-confirmation.tsx
```diff
@@ -0,0 +1,63 @@
...
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "@openstatus/ui";
+
+import { LoadingAnimation } from "@/components/loading-animation";
+import type { MonitorProps } from "../forms/montitor-form";
+
+type FailedPingAlertConfirmationProps = {
+  submit: ({ ...props }: MonitorProps) => void;
```
mxkaske: I don't think we are using it. If not, can be removed.

apps/web/src/components/modals/failed-ping-alert-confirmation.tsx
```diff
@@ -0,0 +1,63 @@
...
+  props: MonitorProps;
+  pingFailed: boolean;
+  setPingFailed: React.Dispatch<React.SetStateAction<boolean>>;
+  loading: boolean;
```
mxkaske: We don't have to pass `loading` here. And I am not sure if it gets correctly updated as the `loading` state is binded on the parent `onSubmit` function which won't be triggered when clicking on "Confirm" here.

Instead, create your own `[isPending, startTransition]` in this file and wrap the function inside it. Similar to the `monitor-form.tsx`.

apps/web/src/components/forms/montitor-form.tsx
```diff
@@ -505,11 +514,16 @@ export function MonitorForm({
           )}
         />
         <div className="sm:col-span-full">
-          {/*
-           * We could think of having a 'double confirmation' one,
-           * to check if the endpoint works and approve afterwards
-           * and confirm anyways even if endpoint failed
-           */}
+          {pingFailed && (
```
mxkaske: You don't need the ternary operation here as we use `open={pingFailed}` to open the Confirmation Modal.

apps/web/src/components/forms/montitor-form.tsx
```diff
@@ -505,11 +514,16 @@ export function MonitorForm({
...
+            <FailedPingAlertConfirmation
+              props={form.getValues()}
+              submit={onSubmit}
+              loading={isPending}
+              pingFailed={pingFailed}
+              setPingFailed={setPingFailed}
```
mxkaske: It's ok. But somehow it feels off to pass the full state to the children. 

How about instead making it more explicit and do something like:
```jsx
open={pingFailed}
toggle={() => setPingFailed(val => !val)}
```

Then, you can pass the `open` prop to the `Alert` and use the `toggle()` to close or open (e.g. `handleSubmit` or `onOpenChange`).

apps/web/src/components/forms/montitor-form.tsx
```diff
@@ -505,11 +514,13 @@ export function MonitorForm({
...
-           * We could think of having a 'double confirmation' one,
-           * to check if the endpoint works and approve afterwards
-           * and confirm anyways even if endpoint failed
-           */}
+          <FailedPingAlertConfirmation
+            monitor={form.getValues()}
+            submit={onSubmit}
+            pingFailed={pingFailed}
+            setPingFailed={setPingFailed}
+            handleDataInsertion={handleDataUpdateOrInsertion}
```
mxkaske: How about

```suggestion
            monitor={form.getValues()}
            onConfirm={handleDataUpdateOrInsertion}
            {...{ pingFailed, setPingFailed }}
```

apps/web/src/components/modals/failed-ping-alert-confirmation.tsx
```diff
@@ -0,0 +1,60 @@
...
+  setPingFailed,
+  monitor,
+}: FailedPingAlertConfirmationProps) => {
+  const [isPending, startTransition] = React.useTransition();
+  const handleSubmit = () => {
+    startTransition(async () => {
+      upsertMonitor(monitor);
+    });
+    setPingFailed(false);
+  };
```
mxkaske: We need to move the `setState` into the transition and add an `await`. Otherwise the modal will close immediately without waiting the save to happen.

You might wanna update the `Props`:

```
onConfirm: (props: MonitorProps) => Promise<void>;
```

```suggestion
  const handleSubmit = () => {
    startTransition(async () => {
      await upsertMonitor(monitor);
      setPingFailed(false);
    });
  };
```

apps/web/src/components/modals/failed-ping-alert-confirmation.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <AlertDialogDescription>
+            The test ping failed. Are you sure you want to continue to save?
+          </AlertDialogDescription>
+        </AlertDialogHeader>
+        <AlertDialogFooter>
+          <AlertDialogCancel>Cancel</AlertDialogCancel>
+          <AlertDialogAction
+            type="submit"
+            disabled={isPending}
+            onClick={handleSubmit}
```
mxkaske: The default behaviour of the `onClick` is to close the modal. We will need to prevent it.

```suggestion
            onClick={(e) => {
              e.preventDefault();
              handleSubmit();
            }}
```

mxkaske (commented):
The only missing part is the modal closing before the submit happens.
To test asynchronous behaviours, we have the helper function `wait`:

```js
import { wait } from "@/lib/utils";

async function handleSubmit() {
  await wait(2000); // 2s
}
```

PR #341, opened 2023-09-30
[OPE-36] Improve Monitor Data Table 

### LINKED ISSUE #194: [OPE-36] Improve Monitor Data Table
### Pagination

Right now, we only have client side pagination. It would be great to have server side pagination to be able to share pages at some point and only load needed data.

### Limits

Same for the limits.

Create page pagination (e.g. `/app/[workspace-slug]/monitors/[id]/data/1?limit=10`, `/app/[workspace-slug]/monitors/[id]/data/2?limit=10`) which then gets offset data from tinybird.

```
SELECT *
FROM your_table
ORDER BY some_column
LIMIT 10 OFFSET 20;
```

Resources:

* [Josh tried coding](https://www.youtube.com/watch?v=SXmni_7B0r4)
* …

<sub>From [SyncLinear.com](https://synclinear.com) | [OPE-36](https://linear.app/openstatus/issue/OPE-36/improve-monitor-data-table)</sub>

PeterKwesiAnsah:
This PR introduces new page query parameters, page, and page_size (which replaces limit) to the response_list pipe.
This fixes #194 allowing the server to handle pagination instead of the client

### DIFF
--- a/apps/web/src/components/data-table/data-table-pagination.tsx
+++ b/apps/web/src/components/data-table/data-table-pagination.tsx
@@ -31,7 +31,7 @@ export function DataTablePagination({ pageCount }: DataTablePaginationProps) {
   const pathname = usePathname();
   const updateSearchParams = useUpdateSearchParams();
 
-  const page_size = searchParams.get("page_size") || "10";
+  const page_size = searchParams.get("page_size") || "20";
 
   return (
     <div className="flex items-center justify-between px-2">
--- a/apps/web/src/components/data-table/data-table-pagination.tsx
+++ b/apps/web/src/components/data-table/data-table-pagination.tsx
@@ -20,12 +20,12 @@ import {
 import useUpdateSearchParams from "@/hooks/use-update-search-params";
 
 interface DataTablePaginationProps {
-  pageCount: number;
+  hasNextPage: boolean;
 }
 
-export function DataTablePagination({ pageCount }: DataTablePaginationProps) {
+export function DataTablePagination({ hasNextPage }: DataTablePaginationProps) {
   const searchParams = useSearchParams();
-  const page = useSearchParams().get("page") || "1";
+  const offset = parseInt(useSearchParams().get("offset") || "0");
 
   const router = useRouter();
   const pathname = usePathname();
@@ -40,7 +40,7 @@ export function DataTablePagination({ pageCount }: DataTablePaginationProps) {
     };
   };
 
-  const page_size = searchParams.get("page_size") || "20";
+  const limit = parseInt(searchParams.get("limit") || "20");
 
   return (
     <div className="flex items-center justify-between px-2">
@@ -49,35 +49,35 @@ export function DataTablePagination({ pageCount }: DataTablePaginationProps) {
         <div className="flex items-center space-x-2">
           <p className="text-sm font-medium">Rows per page</p>
           <Select
-            value={page_size}
+            value={limit + ""}
             onValueChange={(value) => {
               updateSearchParamsWithRouter({
-                page_size: value,
-                page: 1,
+                limit: value,
+                offset: 0,
               })();
             }}
           >
             <SelectTrigger className="h-8 w-[70px]">
-              <SelectValue placeholder={page_size} />
+              <SelectValue placeholder={limit} />
             </SelectTrigger>
             <SelectContent side="top">
-              {[10, 20, 30, 40, 50].map((pageSize) => (
-                <SelectItem key={pageSize} value={`${pageSize}`}>
-                  {pageSize}
+              {[1, 10, 20, 30, 40, 50].map((limit) => (
+                <SelectItem key={limit} value={`${limit}`}>
+                  {limit}
                 </SelectItem>
               ))}
             </SelectContent>
           </Select>
         </div>
-        <div className="flex w-[100px] items-center justify-center text-sm font-medium">
+        {/* <div className="flex w-[100px] items-center justify-center text-sm font-medium">
           Page {page} of {pageCount}
-        </div>
+        </div> */}
         <div className="flex items-center space-x-2">
           <Button
             variant="outline"
             className="hidden h-8 w-8 p-0 lg:flex"
             onClick={updateSearchParamsWithRouter({ page: "1" })}
-            disabled={!(page > "1")}
+            disabled={!offset}
           >
             <span className="sr-only">Go to first page</span>
             <ChevronsLeft className="h-4 w-4" />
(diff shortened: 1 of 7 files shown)


apps/web/src/lib/tb.ts
```diff
@@ -30,6 +33,25 @@ export async function getResponseListData(
   return;
 }
 
+export async function getResponseListDataCount(
```
PeterKwesiAnsah: The Point of this function is to get the total responses in the response list, allowing the calculation of the total page count

packages/tinybird/src/validation.ts
```diff
@@ -69,7 +69,8 @@ export const tbParameterResponseList = z.object({
   monitorId: z.string().default(""), // REMINDER: remove default once alpha
   fromDate: z.number().int().default(0), // always start from a date
   toDate: z.number().int().optional(),
-  limit: z.number().int().optional().default(2500), // one day has 2448 pings (17 (regions) * 6 (per hour) * 24)
+  page: z.number().int().optional().default(0),
+  page_size: z.number().int().optional().default(20), // one day has 2448 pings (17 (regions) * 6 (per hour) * 24)
```
mxkaske: Is it possible to keep `limit` and call the other parameter `offset` instead of `page`? That way, we stay close to the SQL syntax.

apps/web/src/lib/tb.ts
mxkaske: This is a good attempt but I think we should create an entirely new `pipe`. Otherwise, we will process too much data from tinybird and the response time will quickly blow up.

I am thinking out loud here:

The following command **224872** entries. Does it even make sense to know the maximum amount of data and therefore page count? Or would it be ok to navigate via the icons and remove the last icon that fast forwards you to the last page as well as the "X out of Y" number.

```
%
SELECT count(*)
FROM ping_response__v3
WHERE monitorId = {{ String(monitorId, 'openstatusPing') }}
```

CleanShot 2023-10-01 at 10 09 29@2x

mxkaske (commented):
Thanks @PeterKwesiAnsah! Clean code. 

Within the code, a question was raised up. Feel free to give your opinion on that!

Otherwise, would be great to keep the names `limit` and `offset`.

packages/tinybird/src/validation.ts
PeterKwesiAnsah: @mxkaske we sure can do that.

PeterKwesiAnsah: @mxkaske so instead of page and page_size, should we limit and offset ??

apps/web/src/lib/tb.ts
PeterKwesiAnsah: @mxkaske I actually thought about creating a new pipe just for the count. It's a good idea.
Umm, the page count is essential.  I mean without it how will we know the user is on his last page so we disable the next page button?

mxkaske: Once the data received are less than the 'rows per page', we know that it is the last page. We can then disable the 'next page button'.

Supposedly we do have the *224.872** rows and 100 rows per page, we will still have **2.249** pages to go to. As a user of OpenStatus, and an incident occurs, I'm interested in the last pings and don't what happened before. 

I think diagrams are more important to get a better overview going to the first pings.

PeterKwesiAnsah: @mxkaske This will work,which means there's no need to get the pages count

PeterKwesiAnsah: @mxkaske done

packages/tinybird/src/validation.ts
PeterKwesiAnsah: @mxkaske done

apps/web/src/components/data-table/data-table-pagination.tsx
```diff
@@ -15,78 +17,102 @@ import {
...
+interface DataTablePaginationProps {
+  hasNextPage: boolean;
 }
 
-export function DataTablePagination<TData>({
-  table,
-}: DataTablePaginationProps<TData>) {
+export function DataTablePagination({ hasNextPage }: DataTablePaginationProps) {
+  const searchParams = useSearchParams();
+  const offset = parseInt(useSearchParams().get("offset") || "0");
```
mxkaske:
```suggestion
  const offset = parseInt(searchParams.get("offset") || "0");
```

apps/web/src/components/data-table/data-table-pagination.tsx
```diff
@@ -15,78 +17,102 @@ import {
...
+  const router = useRouter();
+  const pathname = usePathname();
+  const updateSearchParams = useUpdateSearchParams();
+
+  const updateSearchParamsWithRouter = (
+    query: Parameters<typeof updateSearchParams>[0],
+  ) => {
+    const updatedSearchParams = updateSearchParams(query);
+    return () => {
+      router.push(`${pathname}/?${updatedSearchParams.toString()}`);
```
mxkaske: I don't think you need the `pathname` here. Also, a `.replace` might be better. Otherwise it will stack a new route on top of the history.

```suggestion
      router.replace(`?${updatedSearchParams.toString()}`);
```

apps/web/src/components/data-table/data-table.tsx
```diff
@@ -110,7 +111,7 @@ export function DataTable<TData, TValue>({
           </TableBody>
         </Table>
       </div>
-      <DataTablePagination table={table} />
```
mxkaske, line 113: Keep the `DataTablePagination` here, you can get the `limit` via the searchParams instead of passing it as `children`.

mxkaske (commented):
Thanks @PeterKwesiAnsah. Left some comments!

apps/web/src/components/data-table/data-table.tsx
PeterKwesiAnsah: @mxkaske I get that, but the reason why I'm using it as a child is to prevent prop-drilling the number of data results.
Remember we need the number of data results to know if there's a next page or not.

apps/web/src/components/data-table/data-table-pagination.tsx
PeterKwesiAnsah: @mxkaske Thanks!. I will look into that.

apps/web/src/components/data-table/data-table.tsx
mxkaske: Yep, gotcha. Let’s keep that for now and iterate later.

apps/web/src/components/data-table/data-table-pagination.tsx
PeterKwesiAnsah: @mxkaske Changes made. Works as expected

apps/web/src/components/data-table/data-table-pagination.tsx
PeterKwesiAnsah: @mxkaske done

mxkaske:
Hey @PeterKwesiAnsah! 

I am very upset to say that we are about to drop the server-side pagination.

We will **only** allow the date range options: 
- Today
- Last 3 days
- Last week
- Last month

And display the information via Chart using tremor.so. At the same time, the idea is to keep the table and the client-side pagination because we will already have the data fetched and available on the client.

If we would do a server-side change (e.g. `limit` or `offset`, we will break the graphs), we would have to create two separate requests to tinybird (one for the chart, one for the table) and therefore double the amount of data fetched.

That's something that came up yesterday while working on charts and I haven't realized it before.

I wonder if I am wrong. So please feel free to leave a comment. I wish you find another issue to contribute and be a contributor!

PeterKwesiAnsah:
@mxkaske  Thanks for Letting me know. You are right, it makes sense.

PR #344, opened 2023-10-01
fix(opengraph): remove unnecessary bracket in metadata

arthurdotwork:
Hey there! 👋 

Just registered and configured my first monitor and while sharing it with the team on Slack, I noticed there was an extra bracket in the generated opengraph metadata. :)

CleanShot 2023-10-01 at 19 06 11@2x

Thanks a lot, that is an useful & great looking tool!


thibaultleouay:
Thanks @arthureichelberger

Issue #346, opened 2023-10-01
Add sitemap generation for docs 


thibaultleouay:
We should also generate a sitemap for the nextra doc site

PR #348, opened 2023-10-02
feat: adding sitemap for docs

Ipriyankrajai:
**Issue Resolution for #346**

In order to resolve issue #346, I've successfully implemented the use of the `next-sitemap` package to generate the sitemap.xml for our documentation website.

**Prerequisites:**

1. Create a `.env` file within the `apps\docs` directory and add the following line:
   ```
   SITE_URL=https://docs.openstatus.dev/
   ```

Now, let's delve into the two methods available for generating the sitemap for our documentation:

**Method 1: Conventional Approach (Recommended by `next-sitemap`)**

This method involves two steps:

   a. First, initiate a build process using the command:
      ```
      pnpm build
      ```
   b. Once the build is completed, execute the following command to generate the sitemap:
      ```
      pnpm postbuild
      ```

**Method 2: Streamlined Approach (Recommended)**

For greater efficiency and to minimize the chance of missing steps, we have devised a single command to handle both build and sitemap generation:

   ```
   pnpm onegobuild
   ```

We highly recommend the second approach (Method 2) as it eliminates the possibility of forgetting to run `pnpm postbuild` when new content is added. This ensures a seamless and automated process for keeping our sitemap up to date.

Feel free to use this approach, and if you have any questions or encounter any issues, please don't hesitate to reach out for assistance.

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -3,6 +3,8 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
+    "onegobuild": "next build && next-sitemap",
+    "postbuild": "next-sitemap",
     "start": "next start"
   },
   "repository": {
@@ -13,6 +15,7 @@
   "dependencies": {
     "@vercel/analytics": "1.0.1",
     "next": "13.4.19",
+    "next-sitemap": "^4.2.3",
     "nextra": "2.13.1",
     "nextra-theme-docs": "2.13.1",
     "react": "18.2.0",
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -3,7 +3,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "onegobuild": "next build && next-sitemap",
+    "build:sitemap": "next build && next-sitemap",
     "postbuild": "next-sitemap",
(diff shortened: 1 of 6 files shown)


apps/docs/package.json
```diff
@@ -3,6 +3,8 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
+    "onegobuild": "next build && next-sitemap",
```
thibaultleouay: might be better to rename it build:sitemap

apps/docs/package.json
```diff
@@ -13,6 +15,7 @@
   "dependencies": {
     "@vercel/analytics": "1.0.1",
     "next": "13.4.19",
+    "next-sitemap": "^4.2.3",
```
thibaultleouay: can you fix the dep (exact version)

apps/docs/package.json
Ipriyankrajai: Sure.

Should I remove `postbuild` script?

thibaultleouay: nope if I think it could be keep

thibaultleouay (approved):
LGTM 🚀

PR #365, opened 2023-10-09
Move to fly

thibaultleouay:
still a wip


apps/server/src/checker/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import { checker, monitor, triggerAlerting, updateMonitorStatus } from "./checker";
+import { payloadSchema } from "./schema";
+
+export const checkerRoute = new Hono();
+
+checkerRoute.post("/checker", async (c) => {
+  const r = new Receiver({
+    currentSigningKey: process.env.QSTASH_CURRENT_SIGNING_KEY || "",
+    nextSigningKey: process.env.QSTASH_NEXT_SIGNING_KEY || "",
+  });
```
mxkaske: Keep it outside. No need to init it on every call.

packages/db/src/schema/monitor.ts
```diff
@@ -13,33 +13,13 @@ import { notificationsToMonitors } from "./notification";
...
-  "hnd1",
-  "iad1",
-  "icn1",
-  "kix1",
-  "lhr1",
-  "pdx1",
-  "sfo1",
-  "sin1",
-  "syd1",
-] as const;
```
mxkaske: Keep them for tinybird. Instead, extend it with the `Flyregion`.

PR #366, opened 2023-10-11
feat: @openstatus/react

mxkaske:


packages/react/src/widget.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export type StatusResponse = { status: Status };
+
+export async function getStatus(slug: string): Promise<StatusResponse> {
+  const res = await fetch(`https://api.openstatus.dev/public/status/${slug}`, {
+    cache: "no-cache",
+  });
+
+  if (res.ok) {
+    const data = (await res.json()) as StatusResponse;
```
thibaultleouay, line 19: No zod ?

packages/react/README.md
thibaultleouay, line 1: Can you add a little bit more content, I think it could be use as npm seo 🤣

packages/react/src/widget.tsx
mxkaske: it will blow the package size

PR #391, opened 2023-10-19
Fix overflow issues on mobile for Status Pages

atridadl:
Issue screenshot:
image

The solution seems to be just moving this to flex box and removing the truncate on the header title and description.


atridadl:
This might solve it? I have to get back to my day job but hopefully this is at least a starting point :)

mxkaske:
Thanks @atridadl for initialising it! I pushed a small change on top of it.

PR #398, opened 2023-10-20
feat: region selections

### LINKED ISSUE #368: Feature request: Support for monitoring using HEAD method
Please consider monitoring using `HEAD` in addition to `GET`/`POST`.

### LINKED ISSUE #369: Bug: Wrong time shown in graph
Looks like the graph is using UTC instead of browser time?

image

### LINKED ISSUE #372: Feture request: Disable/pause monitors
Would be nice to be able to disable/pause monitors instead of deleting them. Could still count against the plans monitor quota, would still be valid in some cases.

### LINKED ISSUE #395: Bug: Unable to go back to using a random region
Maybe I am missing something obvious, but I am unable to go back to a random region after first selecting one. I can only switch to different regions but not unselect it.

mxkaske:
- [x] allow users to unselect specific regions
- [x] allow random regions
- [x] fix time formatting
- [x] remove vercel integration from everywhere
- [x] update faq 
- [x] toggle monitor active in action button
- [x] support HEAD method


apps/web/.env.example
thibaultleouay, line 1: I think we can remove the vercel. Cli version

mxkaske: good point

PR #402, opened 2023-10-21
fix: correct monitor form component name

lsmda:


mxkaske:
Great catch! Thanks @lucasamda

Issue #409, opened 2023-10-24
Bug: Initial ping when creating a monitor seems to not use selected HTTP method


stefanes:
### Preliminary Checks

- [X] I have reviewed the documentation: https://docs.openstatus.dev/

- [X] I have searched for existing issues: https://github.com/openstatusHQ/openstatus/issues

- [X] This issue is not a question, general help request, or anything other than a bug report directly related to OpenStatus. Please ask questions in our Discord community: https://www.openstatus.dev/discord

### Description

**Steps to reproduce:**

1. Create a new monitor for a host that only allows HEAD requests
2. Set the HTTP method to **HEAD** under _HTTP Request Settings_
3. Click **Create**

**Expected behavior:**

Initial ping is performed using the selected method and the monitor is created without any warning displayed to the user.

**Actual behavior:**

The initial ping request seems to not use the method selected under _HTTP Request Settings_, and a warning is displayed to the user:

![image](https://github.com/openstatusHQ/openstatus/assets/5484354/0f0ef087-02ce-4668-a8b2-895e008221b9)

AmoabaKelvin:
Thank you @stefanes for reporting this. Will jump into fixing that ASAP

stefanes:
Same issue seems to be true for the _Test endpoint_ functionality btw:

![image](https://github.com/openstatusHQ/openstatus/assets/5484354/9efda0c7-f46e-4fbb-8ced-85903a9fba5d)

AmoabaKelvin:
Hey @stefanes, issue identified and fixed, should be okay once we merge and deploy 🚀 

Cheers, 🥂

PR #410, opened 2023-10-24
Fix issue when test pinging a monitor endpoint

AmoabaKelvin:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Initially when the `HEAD` request method is selected, the body is still passed in to the fetch call which will cause a failure in the ping.

### Before this PR
Testing monitor endpoints with `HEAD` method no fails

### After this PR
Testing monitor endpoints with `HEAD` method no longer fails

### Related Issue (optional)
#409


thibaultleouay:
🔥 lgtm

PR #422, opened 2023-10-30
Grammar Polishing

diogoribeirodev:
English is not my first language, but i think this way it looks/is written better!


diogoribeirodev:
I changed reason to reasons bc technically there are two, a **lightweight server** and **pricing**

mxkaske:
Thanks @diogoribeirodev!

PR #463, opened 2023-11-10
🚀 SMS notifications 

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

SMS 🔥


packages/notifications/twillio-sms/src/schema/config.ts
```diff
@@ -0,0 +1,6 @@
+import isMobilephone from "validator/lib/isMobilePhone";
+import { z } from "zod";
+
+export const SmsConfigurationSchema = z.object({
+  phoneNumber: z.string().refine(isMobilephone),
```
mxkaske, line 5: Can we use regex instead of adding a deps for that specific use case?

```js
const phoneRegex = new RegExp(
  /^([+]?[\s0-9]+)?(\d{3}|[(]?[0-9]+[)])?([-]?[\s]?[0-9])+$/
);

const schema = z.object({
     phone: z.string().regex(phoneRegex, 'Invalid Number!'),
})
```

PR #472, opened 2023-11-13
feat: tb update schema

mxkaske:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/web/src/components/marketing/cards.tsx
+++ b/apps/web/src/components/marketing/cards.tsx
@@ -20,12 +20,12 @@ export function Cards(props: Feature) {
           return (
             <li key={i}>
               <p className="flex flex-col">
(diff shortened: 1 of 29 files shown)


apps/web/src/components/marketing/cards.tsx
```diff
@@ -20,12 +20,12 @@ export function Cards(props: Feature) {
           return (
             <li key={i}>
               <p className="flex flex-col">
-                <p>
+                <span>
```
mxkaske, line 23: @AmoabaKelvin, fyi: with RSC you get a hydration error when including a `p` inside of a `p`.

PR #547, opened 2023-12-31
feat(checker): http cleanup

arthurdotwork:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

I did a bit of a cleanup on the HTTP scope of the checker by:
- Using Gin as it gives you some help in handling requests & responses — I assume that the tiny advantage on speed is negligible here.
- Introducing the usage of contexts.
- Introducing graceful shutdown with contexts — not used yet in the proper checker but that's the goal of another PR.
- Extracted the request struct in its directory.

### Before this PR

n/a.

### After this PR

n/a.

### Related Issue (optional)

n/a.


thibaultleouay (approved):
🔥

PR #548, opened 2023-12-31
feat(checker): extract checker command

arthurdotwork:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

#547 should be reviewed before.

I am just extracting the main.go file in a cmd/ package so that the top functions like Ping, UpdateStatus ... can live in the checker root package. I also added a missing part on the graceful shutdown: when the channel receives something, it now cancels the context that shutdown the HTTP server.

### Before this PR

n/a.

### After this PR

n/a.

### Related Issue (optional)

n/a.

### DIFF
--- a/apps/checker/Dockerfile
+++ b/apps/checker/Dockerfile
@@ -11,7 +11,7 @@ COPY . .
 
 ARG VERSION
 
-RUN go build -o openstatus-checker .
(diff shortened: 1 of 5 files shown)


thibaultleouay (commented):
🔥

PR #569, opened 2024-01-10
feat(checker): extract tinybird in its own package

arthurdotwork:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Just extracting Tinybird in its own package & adding tests. I used a tip that allow you to hijack what happens inside the httpClient so that you can basically test everything and returns whatever you want. For now, it's fairly simple but it can be improved — _I may even write my own open source package to manage this_.

### Before this PR

n/a.

### After this PR

n/a.

### Related Issue (optional)

n/a.


thibaultleouay:
Merci 🔥

PR #570, opened 2024-01-10
feat(checker): implement backoff retry

arthurdotwork:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Updating some proper backoff retry instead of nested retry inside errors. I also clarified some use cases so the only reasons for the request to be retried inside the backoff would be:
- We were unable to build the request — it is not likely to happen but we still need to handle that use case.
- The server behind the monitor is unreachable — not a timeout or bad error code, just unreachable.

These two use cases will end in an immediate retry inside the first HTTP call to the checker.

Added a few questions (prefixed by `// Q:`) for some use cases as there seems to be a logic before the status update. My take is that the checker itself should be dumb and only execute the check and return the status of the check to the caller which should itself handle the status.

### Before this PR

n/a.

### After this PR

n/a.

### Related Issue (optional)

n/a.

### DIFF
--- a/apps/checker/cmd/main.go
+++ b/apps/checker/cmd/main.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
-	"strconv"
 	"syscall"
 
 	"github.com/gin-gonic/gin"
@@ -16,7 +15,15 @@ import (
 	"github.com/openstatushq/openstatus/apps/checker/pkg/tinybird"
 	"github.com/openstatushq/openstatus/apps/checker/request"
 	"github.com/rs/zerolog/log"
+
+	backoff "github.com/cenkalti/backoff/v4"
 )
+
+type statusCode int
+
+func (s statusCode) IsSuccessful() bool {
+	return s >= 200 && s < 300
+}
 
 func main() {
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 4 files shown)


arthurdotwork:
Review #569 before as the PRs are overlapping.

apps/checker/cmd/main.go
```diff
@@ -48,84 +60,71 @@ func main() {
...
+				CronTimestamp: req.CronTimestamp,
+				Timestamp:     req.CronTimestamp,
+				MonitorID:     req.MonitorID,
+				WorkspaceID:   req.WorkspaceID,
+			}); err != nil {
+				log.Ctx(ctx).Error().Err(err).Msg("failed to send event to tinybird")
+			}
+
+			// If the status was previously active, we update it to error.
+			// Q: Why not always updating the status? My idea is that the checker should be dumb and only check the status and return it.
```
thibaultleouay, line 116: we will 😁 
We are going to create incident for monitors and it will be done here  
The status  is a band aid to manage that 🤣

thibaultleouay:
@arthureichelberger thanks for the PR, 
If you can resolve conflict it would be 🔥

thibaultleouay (approved):
LGTM 🔥

PR #572, opened 2024-01-12
improve checker

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
Let's store more data 

Some timing and some headers


apps/checker/ping.go
```diff
@@ -65,6 +105,16 @@ func Ping(ctx context.Context, client *http.Client, inputData request.CheckerReq
 	}
 	defer response.Body.Close()
 
+	_, err = io.ReadAll(response.Body)
+	if err != nil {
```
arthurdotwork, line 128:
```suggestion
	if _, err := io.ReadAll(response.Body); err != nil {
```

arthurdotwork: & not sure you need to decode the body to get the headers actually given that you don't do anything of the body.

thibaultleouay: The endpoint can return some data, and to make sure the data is processable.

apps/checker/ping.go
```diff
@@ -57,6 +96,7 @@ func Ping(ctx context.Context, client *http.Client, inputData request.CheckerReq
 				Timestamp:   time.Now().UTC().UnixMilli(),
 				URL:         inputData.URL,
 				Message:     fmt.Sprintf("Timeout after %d ms", latency),
+				Timing:      timing,
```
mxkaske: Instead of saving `DnsStart` and `DnsDone`, lets only save the milliseconds needed, e.g. `dns` (will save some space on tinybird that's just the information we need)

Issue #612, opened 2024-01-31
Improve tests coverage for the API server 


thibaultleouay:
### Description

We currently have only a limited set of tests for the API server (`/apps/server/src/v1`) 
We should need to improve it. 
We should have some tests for every endpoints

thibaultleouay:
@skushagra9  If you are interested this one can be for you

skushagra9:
Thanks , I will work on this

skushagra9:
This is solved

PR #613, opened 2024-01-31
Added tests for all the endpoints

skushagra9:
#612 

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/server/src/v1/incidents.test.ts
+++ b/apps/server/src/v1/incidents.test.ts
@@ -3,7 +3,7 @@ import { expect, test } from "bun:test";
 import { api } from ".";
 
 test("GET one Incident", async () => {
-  const res = await api.request("/incident/1", {
+  const res = await api.request("/incident/2", {
     headers: {
       "x-openstatus-key": "1",
     },
@@ -13,6 +13,11 @@ test("GET one Incident", async () => {
     id: 1,
     startedAt: expect.any(String),
     monitorId: 1,
+    acknowledgedAt: "2023-11-08T21:03:13.000Z",
+    resolvedAt: null,
+    resolvedBy: null,
+    acknowledgedBy: null,
+
   });
 });
 
@@ -42,17 +47,20 @@ test("Update an incident ", async () => {
 
 test("Get all Incidents", async () => {
   const res = await api.request("/incident", {
+    method:"GET",
     headers: {
       "x-openstatus-key": "1",
     },
   });
   expect(res.status).toBe(200);
-  expect(await res.json()).toMatchObject({
+  expect((await res.json())).toMatchObject({
+    acknowledgedAt: "2023-11-08T21:03:13.000Z",
+    monitorId: 1,
     id: 1,
     startedAt: expect.any(String),
-    monitorId: 1,
+    resolvedAt: null,
+    resolvedBy: null,
+    acknowledgedBy: null,
   });
 
 });
-
-
--- a/apps/server/src/v1/incidents.test.ts
+++ b/apps/server/src/v1/incidents.test.ts
@@ -39,3 +39,14 @@ test("Update an incident ", async () => {
     acknowledgedBy: null,
   });
 });
+
+test("Get all Incidents", async () => {
+  const res = await api.request("/incident", {
+    headers: {
+      "x-openstatus-key": "1",
+    },
+  });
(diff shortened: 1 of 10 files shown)


apps/server/src/v1/incidents.test.ts
```diff
@@ -39,3 +39,14 @@ test("Update an incident ", async () => {
...
   });
 });
+
+test("Get all Incidents", async () => {
+  const res = await api.request("/incident", {
+    headers: {
+      "x-openstatus-key": "1",
+    },
+  });
+  expect(res.status).toBe(200);
```
thibaultleouay, line 59: I think for this one we could also check it returns an array that contains two element

skushagra9:
get monitor daily summary route's test is missing , I will add it

packages/db/src/schema/monitors/constants.ts
```diff
@@ -33,7 +33,7 @@ export const monitorPeriodicity = [
   "1h",
   "other",
 ] as const;
-export const monitorMethods = ["GET", "POST", "HEAD"] as const;
+export const monitorMethods = ["GET", "POST", "HEAD", "PUT"] as const;
```
thibaultleouay: we don't allow PUT here I think we would need to do a DB migration let's remove it now and let's fix it later with #619

apps/server/src/v1/monitor.ts
```diff
@@ -533,13 +534,16 @@ monitorApi.openapi(getMonitorStats, async (c) => {
...
+
   if (workspaceId !== _monitor.workspaceId)
     return c.jsonT({ code: 401, message: "Unauthorized" });
+  console.log("here")
 
   const cache = await redis.get<z.infer<typeof dailyStatsSchemaArray>>(
     `${monitorId}-daily-stats`,
   );
-
+    console.log("now")
```
thibaultleouay: Can you remove it please?

apps/server/src/v1/monitor.test.ts
```diff
@@ -69,3 +80,206 @@ test("Create a monitor", async () => {
...
+    },
+  });
+  expect(res.status).toBe(200);
+
+  expect(await res.json()).toMatchObject({
+    message: "Deleted"
+  });
+});
+
+// test("Get monitor daily Summary", async () => {
```
thibaultleouay, line 267: We can prefix test with todo 
```
test.todo
``` 
So we get a reminder to do it

apps/server/src/v1/monitor.test.ts
```diff
@@ -69,3 +80,206 @@ test("Create a monitor", async () => {
...
+  expect(res.status).toBe(401);
+})
+test("Update a Monitor-403", async () => {
+  const data = {
+    periodicity: 9,//not passing correct value returns 403
+    url: "https://www.openstatus.dev",
+    name: "OpenStatus",
+    description: "OpenStatus website",
+    regions: ["ams"],
+    method: "PUT",
```
thibaultleouay: We don't have put can you use GET instead at the moment please

apps/server/src/v1/monitor.test.ts
```diff
@@ -69,3 +80,206 @@ test("Create a monitor", async () => {
...
+    headers: {
+      "x-openstatus-key": "1",
+      "content-type": "application/json",
+    },
+    body: JSON.stringify(data),
+  });
+  expect(res.status).toBe(200);
+
+  expect(await res.json()).toMatchObject({
+    id: expect.any(Number),
```
thibaultleouay: the id should be 1

apps/server/src/v1/monitor.test.ts
thibaultleouay, line 1: In the test if you can rename the name of the test with the explanation why it should failed it will be better to understand them later
e.g. 
Create a monitor -401 -> Create a monitor without auth key should return 401

If you can do that for all the new tests it's perfect

thibaultleouay (commented):
Few comments but it's good

apps/server/src/v1/monitor.ts
thibaultleouay, line 1: Can you remove the console.log please ?

thibaultleouay (approved):
🔥 LGTM

Issue #623, opened 2024-02-05
Make sur the API return  or accept the date in ISO 8601 format


thibaultleouay:
### Description

We should make sure all the returned  or accepted date are in ISO8601 format 
We should also update the zod open api to mention the date should be in this format

PR #625, opened 2024-02-05
fixed tests-2

skushagra9:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### Related Issue (optional)

#623 
removed the 3 entry for statusReport as not needed
added the iso8601 validation to dates

### DIFF
--- a/apps/server/src/v1/monitor.ts
+++ b/apps/server/src/v1/monitor.ts
@@ -31,6 +31,14 @@ const ParamsSchema = z.object({
     }),
 });
 
+const isoDate = z
+  .preprocess((val) => {
+    if (val) {
+      return new Date(String(val)).toISOString();
+    }
+    return new Date().toISOString();
+  }, z.string());
+
 export const periodicityEnum = z.enum(monitorPeriodicity);
 
 const regionInput = z.array(z.enum(flyRegions)).transform((val) => String(val));
@@ -482,7 +490,7 @@ const dailyStatsSchema = z.object({
     .int()
     .openapi({ description: "The total number of request" }),
   avgLatency: z.number().int().openapi({ description: "The average latency" }),
-  day: z.string().openapi({ description: "the date of the event" }),
+  day: isoDate,
 });
 
 const dailyStatsSchemaArray = z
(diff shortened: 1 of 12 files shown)


skushagra9:
adding for monitor

apps/server/src/v1/statusReportUpdate.ts
```diff
@@ -37,10 +37,15 @@ export const statusUpdateSchema = z.object({
...
-    .preprocess((val) => String(val), z.string())
-    .openapi({
-      description: "The date of the update in ISO 8601 format",
-    }),
+  .preprocess((val) => {
+    if (val) {
+      return new Date(String(val)).toISOString();
+    }
+    return new Date().toISOString();
+  }, z.string())
```
thibaultleouay: We should extract that to 
 
```
const isoDate = z
  .preprocess((val) => {
    if (val) {
      return new Date(String(val)).toISOString();
    }
    return new Date().toISOString();
  }, z.string())
  ```
  Thus if we have a single source of truth

Comment (skushagra9) [reply apps/server/src/v1/statusReportUpdate.ts]: sure

Comment (skushagra9) [reply apps/server/src/v1/statusReportUpdate.ts]: should I store it in some other file , as its used in many places ??

File: apps/server/src/v1/statusReport.test.ts [hunk 2]
```diff
@@ -1,7 +1,8 @@
 import { expect, test } from "bun:test";
-
 import { api } from ".";
 
+const iso8601Regex: RegExp = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?Z$/;
```
thibaultleouay: We should probably export in a file test-utils.ts

apps/server/src/v1/monitor.ts
```diff
@@ -31,6 +31,14 @@ const ParamsSchema = z.object({
     }),
 });
 
+const isoDate = z
+  .preprocess((val) => {
```
thibaultleouay: this should be exported in a `utils.ts` and import it where ever we need it

skushagra9: sure

skushagra9:
is it working now ?

thibaultleouay (approved):
🔥 Perfect

PR #655, opened 2024-02-13
 🔥auto resolve incident

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Auto resolve incidents !


apps/server/src/checker/index.ts
mxkaske, line 1: Feels like we are doing multiple same queries. Be it getting `currentMonitor` or `incident`.

thibaultleouay: WDYM ?

PR #681, opened 2024-02-21
Add an additional ping endpoint for separate testing

cramforce:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/web/src/app/api/ping/test/route.ts
+++ b/apps/web/src/app/api/ping/test/route.ts
@@ -1,0 +1,14 @@
+import { NextResponse } from "next/server";
+
+export const dynamic = "force-dynamic";
+
+export const maxDuration = 31; // to trick and not using the same function as the other ping route
+
+export async function GET() {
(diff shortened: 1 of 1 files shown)


elie222:
What's this trick?

```ts
export const maxDuration = 31; // to trick and not using the same function as the other ping route
```

skushagra9:
> What's this trick?
> 
> ```ts
> export const maxDuration = 31; // to trick and not using the same function as the other ping route
> ```

In simple terms, vercel doesn't allows the the request to take more time than 15s , so by specifying this it allows more time for the request

cramforce:
The trick is that functions that have different `maxDuration` get bundled into different physical functions. Since we were playing with cold boots, we wanted to make sure that one invocation doesn't warm up the other ones.

In this case the value doesn't matter, it just need to be unique

PR #719, opened 2024-03-18
🔥 rss feed

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


tarasglek:
I think you need to add some head tags to make this discoverable by feed readers and/or a link with rss to blog

thomasbnt:
> I think you need to add some head tags to make this discoverable by feed readers and/or a link with rss to blog

It's a good idea, better accessibility, to know if the website have an RSS feed. 

```html
<link rel="alternate" type="application/rss+xml" href="/blog/feed.xml" />
```

PR #720, opened 2024-03-20
🔥 status code assertion

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


packages/db/src/schema/monitors/monitor.ts
```diff
@@ -36,6 +37,8 @@ export const monitor = sqliteTable("monitor", {
   name: text("name", { length: 256 }).default("").notNull(),
   description: text("description").default("").notNull(),
 
+  statusCode: text("status_code").default("200"),
```
mxkaske, line 40: We should allow `null` and **not** default to `200` as we might break currently used monitors.

PR #727, opened 2024-03-23
chore: assertions package

mxkaske:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


apps/checker/cmd/main.go
```diff
@@ -93,12 +103,58 @@ func main() {
...
+						fmt.Println("⚠️ Not Handled assertion type", assert.AssertionType)
+					}
+				}
+			} else {
+				isSuccessfull = statusCode.IsSuccessful()
+			}
+
 			// let's retry at least once if the status code is not successful.
-			if !statusCode.IsSuccessful() && called < 2 {
+			if !isSuccessfull && called < 2 {
```
mxkaske, line 145: Should we **only** consider retry when status code is wrong, or the whole assertions?

apps/checker/cmd/main.go
```diff
@@ -136,6 +193,8 @@ func main() {
 				Timestamp:     req.CronTimestamp,
 				MonitorID:     req.MonitorID,
 				WorkspaceID:   req.WorkspaceID,
+				Error:         1,
```
mxkaske, line 196: Are you sure about that?

apps/checker/cmd/main.go
thibaultleouay: Whole assertion

To avoid false positive

apps/checker/cmd/main.go
thibaultleouay: Yep this happens when we have an error at the third retry

PR #734, opened 2024-04-01
feat: add validation for invalid test URLs

yassinrais:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

Hey!

I noticed that you can use links with different variations of the domain name and protocol. Here are some examples:

| Link | Explanation |
| --- | --- |
| <https://www.openstatus.dev/api/checker/test/> | Includes "www" and extra "/" |
| <https://openstatus.dev/api/checker/test/> | Does not include "www" |
| <http://www.openstatus.dev/api/checker/test/> | Uses HTTP instead of HTTPS |
| <http://openstatus.dev/api/checker/test/> | Uses HTTP instead of HTTPS and does not include "www" |

As we are only checking the URL string, we do not check the redirection of the links or the protocols. We only check the incoming URL.

It might be better if you parse the URL using `new URL()` and check if the origin is not the same as one of the hostnames configured in production (or the current host name included in the headers). 
Alternatively, if you want to continue allowing the use of your domain name, you can use `new URL(link).pathname` to check if the pathname is not strictly equal to the current route.

Hope this helps! Let me know if you have any questions 😁

Related: #732

### DIFF
--- a/apps/web/src/app/api/checker/test/route.ts
+++ b/apps/web/src/app/api/checker/test/route.ts
@@ -5,6 +5,7 @@ import { monitorFlyRegionSchema } from "@openstatus/db/src/schema";
 
 import { checkRegion } from "@/app/play/checker/[id]/utils"; // TODO: move into a shared package
 import { payloadSchema } from "../schema";
+import { isAnInvalidTestUrl } from "../utils";
 
 export const runtime = "edge";
 export const preferredRegion = "auto";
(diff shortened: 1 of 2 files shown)


mxkaske:
Thanks @yassinrais, appreciate the improvement! ❤️

PR #741, opened 2024-04-04
[FIX] Get all Page endpoint to return expected array as response

Nabhag8848:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

- Get All Page https://docs.openstatus.dev/api-reference/page/get-page is expected to return an Array of Pages instead of object. This is breaking all Endpoints of Page in production.

### Before this PR
- before it was giving Internal Server Error (Server was Crashing)

### After this PR

StatusPage

### DIFF
--- a/apps/server/src/v1/page.test.ts
+++ b/apps/server/src/v1/page.test.ts
@@ -1,7 +1,6 @@
 import { expect, test } from "bun:test";
 
 import { api } from ".";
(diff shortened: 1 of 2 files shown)


Nabhag8848:
@mxkaske @thibaultleouay

mxkaske:
Thanks for the fix!

PR #801, opened 2024-05-04
migrate: Clerk to Next Auth

mxkaske:
We have added `onUpdate: "cascade"` to the `user.id` to manually change the update the id when testing locally (and automatically update foreign keys as well)

> Manually to do the migration: change the `id` of the `auth_user` to the one used in the `user` db table

Check vercel for the env vars:

```bash
AUTH_SECRET=
AUTH_DRIZZLE_URL=http://127.0.0.1:8080
AUTH_GITHUB_ID=
AUTH_GITHUB_SECRET=
# missing secrets - google oauth not working
AUTH_GOOGLE_ID=
AUTH_GOOGLE_SECRET=
```


packages/db/src/schema/users/user.ts
```diff
@@ -5,18 +5,25 @@ import {
...
   firstName: text("first_name").default(""),
   lastName: text("last_name").default(""),
-  email: text("email").default(""),
   photoUrl: text("photo_url").default(""),
 
+  // next-auth fields
+  name: text("name"),
+  email: text("email").default(""),
+  emailVerified: integer("emailVerified", { mode: "timestamp_ms" }),
+  image: text("image"),
```
mxkaske: if we don't use it, lets remove `image`!

apps/web/src/next-auth.d.ts
mxkaske, line 1: Would be great to have the proper types!

PR #848, opened 2024-06-01
refactor(api): improve api and error handling

### LINKED ISSUE #855: Improve Error Handling (incl. toast notifications)
Currently, we provide very little information to the user whenever something goes wrong. Mostly, some generic phrase keys that can be found in [@lib/toast](https://github.com/openstatusHQ/openstatus/blob/main/apps/web/src/lib/toast.tsx).

The goal is to have concrete informations and possibly links to the _Docs_ that we can forward to (see [Nextjs](https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout) example).

We should create a package (`@openstatus/error`) that can be used on all JS topics, like the API, web, and trpc or react-form-hook (ZodSchema).

The `BaseError` could be:

```ts
type BaseError {
  id: string // unique identify for logging e.g.
  code: ErrorCode, // enum of e.g. "UNAUTHORIZED" | "USAGE_EXCEEDED" ... 
  message: string, // description, can be used for toast notifications or logger
  metadata: Record<string, unknown> // e.g. cause,...
}
```

We could use the error `code` to forward the user within the docs like https://docs.openstatus.dev/error-handling#USAGE_EXCEEDED.

### API server

In our `server` api hono application, we can use `app.onError(handleError)` (see [docs](https://hono.dev/api/exception#handling-httpexception)) to handle thrown exceptions.

AND we can use `defaultHook: handleZodError` for the zod-openapi plugin (see [docs](https://github.com/honojs/middleware/tree/main/packages/zod-openapi#a-dry-approach-to-handling-validation-errors)).

### TRPC

Could be either in the `@openstatus/api` (bad naming as it is about trpc) or within the `trpc/edge/[trpc]/route.ts` file via:

- All error, before getting send to the client will go through `onError` method (see [docs](https://trpc.io/docs/server/error-handling#handling-errors))
-

mxkaske:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


apps/docs/api-reference/errors/code/CONFLICT.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: CONFLIC
```
thibaultleouay: Conflict

packages/db/package.json
```diff
@@ -10,7 +10,8 @@
     "studio": "drizzle-kit studio",
     "env": "bun env.ts",
     "seed": "bun src/seed.mts",
-    "dev": "turso dev --db-file ../../openstatus-dev.db"
+    "dev": "turso dev --db-file ../../openstatus-dev.db",
+    "reset": "rm ./openstatus.db"
```
thibaultleouay: are you sure it's the same file

mxkaske: Lets remove it entirely.

PR #870, opened 2024-06-11
Add run api

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


apps/checker/cmd/main.go
```diff
@@ -61,7 +60,7 @@ func main() {
 	router := gin.New()
 	router.POST("/checker", func(c *gin.Context) {
 		ctx := c.Request.Context()
-
+		dataSourceName := "ping_response__v8"
```
mxkaske, line 63: When do we store it in `ping_response`, when in `check_response`? We are using that endpoint for the tests/playground as well.

apps/checker/ping.go
```diff
@@ -49,14 +49,15 @@ type Timing struct {
...
-	Headers map[string]string `json:"headers,omitempty"`
-	Time    int64             `json:"time"`
-	Timing  Timing            `json:"timing"`
-	Error   string            `json:"error,omitempty"`
-	Tags    []string          `json:"tags,omitempty"`
+	Status    int               `json:"status,omitempty"`
+	Latency   int64             `json:"latency"`
+	Body      string            `json:"body,omitempty"`
+	Headers   map[string]string `json:"headers,omitempty"`
+	Timestamp int64             `json:"timestamp"`
```
mxkaske: Can we keep `Time`? We already store all the data like that. It will break a lot I think.

We use the `zod` schema with 'time' on multiple places like (`/apps/web/src/components/ping-response-analysis/utils.ts`)

apps/checker/ping.go
```diff
@@ -217,10 +218,10 @@ func SinglePing(ctx context.Context, client *http.Client, inputData request.Ping
 		var urlErr *url.Error
 		if errors.As(err, &urlErr) && urlErr.Timeout() {
 			return Response{
-				Latency: latency,
-				Timing:  timing,
-				Time:    start.UTC().UnixMilli(),
-				Error:   fmt.Sprintf("Timeout after %d ms", latency),
+				Latency:   latency,
+				Timing:    timing,
+				Timestamp: start.UTC().UnixMilli(),
```
mxkaske: Same here with `Time`

apps/checker/ping.go
```diff
@@ -235,10 +236,10 @@ func SinglePing(ctx context.Context, client *http.Client, inputData request.Ping
 	}
 
 	return Response{
-		Time:    start.UTC().UnixMilli(),
-		Status:  res.StatusCode,
-		Headers: headers,
-		Timing:  timing,
-		Latency: latency,
+		Timestamp: start.UTC().UnixMilli(),
```
mxkaske: And here.

apps/server/src/v1/check/post.ts
```diff
@@ -0,0 +1,226 @@
...
+      }
+
+      const allResults = await Promise.allSettled(currentFetch);
+      result.push(...allResults);
+    }
+    console.log(result);
+    const filteredResult = result.filter((r) => r.status === "fulfilled");
+    const fulfilledRequest = [];
+    for await (const r of filteredResult) {
+      if (r.status !== "fulfilled") throw new Error("No value");
```
mxkaske, line 100: Do we want to error out here?

apps/server/src/v1/check/post.ts
```diff
@@ -0,0 +1,226 @@
...
+
+      const json = await r.value.json();
+      console.log(ResponseSchema.safeParse(json));
+      fulfilledRequest.push(ResponseSchema.parse(json));
+    }
+
+    let aggregatedResponse = null;
+    if (aggregated) {
+      // This is ugly
```
mxkaske, line 109: Haha oui

apps/server/src/v1/monitors/put.ts
mxkaske, line 1: If we support assertions in `post`,  can we include them in `put`?

apps/server/src/v1/monitors/post.ts
```diff
@@ -67,18 +74,33 @@ export function registerPostMonitor(api: typeof monitorsApi) {
...
+          assert.push(new HeaderAssertion({ ...a, version: "v1" }));
+        }
+        if (a.type === "textBody") {
+          assert.push(new TextBodyAssertion({ ...a, version: "v1" }));
+        }
+        if (a.type === "status") {
+          assert.push(new StatusAssertion({ ...a, version: "v1" }));
+        }
+      }
+    }
```
mxkaske: Lets just use the `import {deserialize} from '@openstatus/assertions'` function on that. It already does what we need.

apps/server/src/v1/monitors/schema.ts
```diff
@@ -6,6 +6,65 @@ import {
...
+
+//   Not used yet
+const _jsonBodyAssertion = z.object({
+  type: z.literal("jsonBody"),
+  path: z.string(), // https://www.npmjs.com/package/jsonpath-plus
+  compare: stringCompare,
+  target: z.string(),
+});
+
+export const assertion = z.discriminatedUnion("type", [
```
mxkaske, line 62: is it a `discriminatedUnion` or a simple `union`?

packages/db/drizzle/0032_hot_swordsman.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE `check` (
```
mxkaske, line 1: Are we storing the checks in the db?

apps/checker/cmd/main.go
thibaultleouay: When we are using the checker endpoint we are using ping_response__v8 
Now when we are using the single api check we are using check_response

packages/db/drizzle/0032_hot_swordsman.sql
thibaultleouay: Yes but not doing anything with them yet

apps/server/src/v1/check/post.ts
thibaultleouay: that should now happened but it's a ts issue because i have filtered the array just 2 lines above

apps/server/src/v1/monitors/post.ts
thibaultleouay: no because we need to add the version :(

thibaultleouay: and assertion from the payload is already an assertion but we don't push the version

apps/server/src/v1/monitors/put.ts
thibaultleouay: ✅

packages/db/drizzle/0032_hot_swordsman.sql
thibaultleouay: Yep we are storing the request on the db not doing anything with it though 
but storing some data,

PR #916, opened 2024-07-05
Fix: issue with cmd + s on status-page/new form

### LINKED ISSUE #914: When saving  monitor, using the keyboard shortcut Cmd + S it should run the checks

_No response_

### Description

API is a reserved slug, so you cannot use it for your status page. However, if you save your status page using the keyboard shortcut Cmd + S, you can still save the page with that slug.

### Environment

_No response_

Ipriyankrajai:
## Type of change

- [x] 🐛 Bug fix

## Description
Moved checkUniqueSlug funtion to onSubmit  function.

Fixes #914 

### Before this PR
There was no validation for api on status page when you  use cmd + s to save

### After this PR
It will check for validation even if you use cmd + s to save

### DIFF
--- a/apps/web/src/components/forms/status-page/form.tsx
+++ b/apps/web/src/components/forms/status-page/form.tsx
@@ -115,35 +115,41 @@ export function StatusPageForm({
   }, [watchTitle, form, defaultValues?.title]);
 
   const onSubmit = async ({ ...props }: InsertPage) => {
-    startTransition(async () => {
-      try {
-        if (defaultValues) {
-          await api.page.update.mutate(props);
-        } else {
-          const page = await api.page.create.mutate(props);
-          const id = page?.id || null;
-          router.replace(`?${updateSearchParams({ id })}`); // to stay on same page and enable 'Advanced' tab
-        }
+    const isUnique = await checkUniqueSlug();
+    if (!isUnique) {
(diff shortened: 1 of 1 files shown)


apps/web/src/components/forms/status-page/form.tsx
```diff
@@ -115,35 +115,41 @@ export function StatusPageForm({
...
-        } else {
-          const page = await api.page.create.mutate(props);
-          const id = page?.id || null;
-          router.replace(`?${updateSearchParams({ id })}`); // to stay on same page and enable 'Advanced' tab
-        }
+    const isUnique = await checkUniqueSlug();
+    if (!isUnique) {
+      // the user will already have the "error" message - we include a toast as well
+      toastAction("unique-slug");
+    } else {
```
mxkaske: Thanks @Ipriyankrajai! One thing: can we move the async call within the `startTransition` so that the pending state gets fired immediately?

mxkaske (approved):
Thanks @Ipriyankrajai!

apps/web/src/components/forms/status-page/form.tsx
Ipriyankrajai: Sure! I was thinking that only

thibaultleouay:
Thanks again  @Ipriyankrajai

PR #921, opened 2024-07-10
Feat: adding clone action to motion page

### LINKED ISSUE #919: Cloning monitors
### Description

We should be able to 'create' or 'clone' a monitor to take all the settings with us (e.g. timeout, assertions, status page relationship, monitor tag relationship,...)

Include a new Dropdown Item

CleanShot 2024-07-08 at 20 38 48@2x

Ipriyankrajai:
Fixes #919
Adding clone action to motion page

- [x] Add `Clone` to DataTableRowActions
- [x] Create a function to clone of monitor
- [x] Redirect user to `monitor/[id]/edit`
- [x] Manage searchParams for `/monitor/[id]/edit?active=true`
- [x] Disable clone action button if limit is hit for the user
- [x] Clone Tags and notifications as well


Ipriyankrajai:
Hi @elie222,

What are your thoughts on the recent changes in [onClone](https://github.com/openstatusHQ/openstatus/pull/921/commits/852696409ac832c66f7c4f65bf17971109dea2ba)? Would you prefer that I create a separate action for this? If so, could you please let me know which folder it should be placed in?

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
```diff
@@ -87,6 +89,75 @@ export function DataTableRowActions<TData>({
   }
   async function onClone() {
```
Ipriyankrajai, line 87: Hi @elie222,

What are your thoughts on the recent changes in onClone? Would you prefer that I create a separate action for this? If so, could you please let me know which folder it should be placed in?

mxkaske: Might got the wrong username. Hey Elie! 👋

Let's create a new `trpc` mutation function [here](https://github.com/openstatusHQ/openstatus/blob/main/packages/api/src/router/monitor.ts).

Ideally, we have your logic to clone a monitor there.

```ts
  cloneMonitorById: protectedProcedure
    .input(z.object({ id: z.number() }))
    .query(async (opts) => {
    ...
  }),
```

Edit: to then use it as `api.monitor.cloneMonitorById.mutate({ id })`

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
```diff
@@ -85,6 +87,78 @@ export function DataTableRowActions<TData>({
...
+          // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+          headerAssertions: _assertions.filter(
+            (a) => a.type === "header"
+          ) as any, // TS considers a.type === "status"
+
+          degradedAfter: data?.degradedAfter,
+          timeout: data?.timeout || 45000,
+        };
+
+        const result = await api.monitor.create.mutate(defaultValues);
```
mxkaske: I think we could do something like below to avoid mapping every prop within the `defaultValues`
```suggestion
        const result = await api.monitor.create.mutate({ ...data, active: false });
```

Ipriyankrajai: Thanks, this helps @mxkaske

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
Ipriyankrajai: Hello @mxkaske ,
I came up with this approach. What are your thoughts on this?

mxkaske: clean

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
```diff
@@ -85,6 +84,35 @@ export function DataTableRowActions<TData>({
...
+          id: undefined,
+          updatedAt: undefined,
+          createdAt: undefined,
+        };
+
+        const createdCloneMonitorData = await api.monitor.create.mutate({
+          ...cloneMonitorData,
+        });
+
+        router.push(`./monitors/${createdCloneMonitorData.id}/edit`);
```
mxkaske:
```suggestion
        router.push(`./monitors/${createdCloneMonitorData.id}/edit`);
        toast.messsage("Monitor cloned!")
```

lets add a toast message to it!

apps/web/src/app/app/[workspaceSlug]/(dashboard)/monitors/[id]/edit/page.tsx
```diff
@@ -30,6 +31,8 @@ export default async function EditPage({
 
   // default is request
   const search = searchParamsSchema.safeParse(searchParams);
+  const active =
+    search.success && search?.data?.active === "true" ? true : monitor.active;
```
Comment (mxkaske) [inline apps/web/src/app/app/[workspaceSlug]/(dashboard)/monitors/[id]/edit/page.tsx]: I think we can use `zod` in a way to transform `active` to a boolean.

Comment (Ipriyankrajai) [reply apps/web/src/app/app/[workspaceSlug]/(dashboard)/monitors/[id]/edit/page.tsx]: Hey @mxkaske ,
Can you check now?

Ipriyankrajai:
Hello @mxkaske,

Could you please review this approach?

packages/api/src/router/monitor.ts
```diff
@@ -741,4 +741,35 @@ export const monitorRouter = createTRPCRouter({
 
     return monitorNumbers >= monitorLimit;
   }),
+  getMonitorStatusTagNotificationById: protectedProcedure
```
mxkaske: Let's call it `getMonitorRelationsById`

packages/api/src/router/monitor.ts
```diff
@@ -741,4 +741,35 @@ export const monitorRouter = createTRPCRouter({
...
+
+
+      const parsedMonitorNotification = monitorNotification.map(({notificationId})=>notificationId)
+      const parsedPages = pages.map(val=>val.pageId)
+      const parsedTags = tags.map(({monitorTagId})=>monitorTagId)
+
+      return {
+        selectedNotification:parsedMonitorNotification,
+        selectedPages:parsedPages,
+        selectedTags:parsedTags
```
mxkaske: A little bit more explicit

```suggestion
        notificationIds: parsedMonitorNotification,
        pageIds: parsedPages,
        monitorTagIds: parsedTags
```

mxkaske (commented):
Logic is fine!

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
Ipriyankrajai: Done!

apps/web/src/app/app/[workspaceSlug]/(dashboard)/monitors/[id]/edit/page.tsx
Comment (Ipriyankrajai) [reply apps/web/src/app/app/[workspaceSlug]/(dashboard)/monitors/[id]/edit/page.tsx]: Done!

packages/api/src/router/monitor.ts
Ipriyankrajai: Done!

packages/api/src/router/monitor.ts
Ipriyankrajai: Done!

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
```diff
@@ -85,6 +86,41 @@ export function DataTableRowActions<TData>({
       }
     });
   }
+  async function onClone() {
+    startTransition(async () => {
+      try {
+        const id = monitor.id;
+        if (!id) return;
+
+        const selectedMonitordata = await api.monitor.getMonitorById.query({
```
mxkaske:
```suggestion
        const selectedMonitorData = await api.monitor.getMonitorById.query({
```

mxkaske (commented):
All fine, works like a charm! 

Two more UI/UX thoughts:

- rename the `monitor.name` and append "copy" at the end when cloning
- add a dialog to confirm the clone (? - not sure on that one though)

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
Ipriyankrajai: Done!

Ipriyankrajai:
Hey @mxkaske ,
Should i append ` - copy` or just `copy` ?
I think we should go with ` - copy` 


image

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
```diff
@@ -37,13 +38,12 @@ interface DataTableRowActionsProps<TData> {
...
-    .object({ monitor: selectMonitorSchema,
-      isLimitReached: z.boolean()
-     })
+  const { monitor, isLimitReached } = z
+    .object({ monitor: selectMonitorSchema, isLimitReached: z.boolean() })
     .parse(row.original);
   const router = useRouter();
   const [alertOpen, setAlertOpen] = React.useState(false);
+  const [alertType, setAlertType] = React.useState<"delete" | "clone" | "">("");
   const [isPending, startTransition] = React.useTransition();
```
mxkaske: I wonder if we can keep a single `useState`

```ts
const [alertType, setAlertType] = React.useState<"clone" | "delete" | undefined>();
const alertOpen = React.useMemo(() => !alertType, [alertType]); // true if "clone" or "delete"
```

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
```diff
@@ -156,21 +171,38 @@ export function DataTableRowActions<TData>({
...
+              else if (alertType === "clone") onClone();
             }}
             disabled={isPending}
-            className="bg-destructive text-destructive-foreground hover:bg-destructive/90"
+            className={cn({
+              "bg-destructive text-destructive-foreground hover:bg-destructive/90":
+                alertType === "delete",
+              "bg-primary text-primary-foreground hover:bg-primary/90":
+                alertType === "clone",
+            })}
```
mxkaske: How about using `asChild` and use the

```ts
<Button variant={x ? "default" : "destructive"}>...</Button>
```

mxkaske (commented):
Maybe we can reduce all the ternary operations by declaring a value like:

```ts
const alerts = {
  clone: {
    action: {
       variant: "default",
       text: "Clone"
    },
    dialog: {
       title: "Title of the Dialog",
       description: "Description..."
    }
  },
  delete: { ... }
} as const
```

to access the values via `alerts[type].action.text` e.g.

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
Ipriyankrajai: Sure

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
Ipriyankrajai: Done

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
Ipriyankrajai: I think we will need this state as we need to pass it to ` <AlertDialog open={alertOpen} onOpenChange={(value) => setAlertOpen(value)}>`

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
Ipriyankrajai: @mxkaske , For some reason  `variant={alerts[alertType].action.variant}` is not working so using ```className={cn(buttonVariants({ variant: alerts[alertType].action.variant }))}```  instead

apps/web/src/components/data-table/monitor/data-table-row-actions.tsx
```diff
@@ -28,22 +29,47 @@ import {
...
-  const { monitor } = z
-    .object({ monitor: selectMonitorSchema })
+  const { monitor, isLimitReached } = z
+    .object({ monitor: selectMonitorSchema, isLimitReached: z.boolean() })
     .parse(row.original);
   const router = useRouter();
-  const [alertOpen, setAlertOpen] = React.useState(false);
-  const [isPending, startTransition] = React.useTransition();
+  const [alertOpen, setAlertOpen] = useState(false);
+  const [alertType, setAlertType] = useState<"delete" | "clone">("delete");
```
mxkaske: Let's use a single state.

```suggestion
  const [alertType, setAlertType] = useState<"delete" | "clone"| undefined>();
  const alertOpen = useMemo(() => !alertType, [alertType])
```

If the type is `undefined`, the open value will be `false`.

That will reduce the redundant state an avoid using `setAlertOpen`.

To reset, we can use `setAlertType(undefined)`.

Ipriyankrajai: Hey @mxkaske ,
But how will me manage `onOpenChnage`  in `<AlertDialog open={alertOpen} onOpenChange={(value) => setAlertOpen(value)}>`

Ipriyankrajai: We don't even need  `const alertOpen = useMemo(() => !alertType, [alertType])` we can use `!!alertType` But again how will we manage `onOpenChange`?

PR #932, opened 2024-07-13
💸 pricing update

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


apps/web/src/app/app/[workspaceSlug]/(dashboard)/notifications/(overview)/_components/channel-table.tsx
```diff
@@ -11,8 +10,8 @@ interface ChannelTable {
 }
 
 export default function ChannelTable({ workspace, disabled }: ChannelTable) {
-  const isPagerDutyAllowed = getLimit(workspace.plan, "pagerduty");
```
Comment (mxkaske) [inline apps/web/src/app/app/[workspaceSlug]/(dashboard)/notifications/(overview)/_components/channel-table.tsx:14]: We should keep the `getLimit` extraction.

That way, if the `workspace.limits = "{}"` we use the default config that we have in `@openstatus/plan` - makes it more dynamic and we only update the database workspaces of paid users.

It gives us a layer of flexibility that we will loose otherwise.

packages/db/src/schema/workspaces/validation.ts
```diff
@@ -3,11 +3,42 @@ import { z } from "zod";
...
 import { workspace } from "./workspace";
+import {
+  monitorFlyRegionSchema,
+  monitorPeriodicitySchema,
+} from "../shared/shared";
 
 export const workspacePlanSchema = z.enum(workspacePlans);
 export const workspaceRoleSchema = z.enum(workspaceRole);
 
+const limitsV1 = z.object({
```
mxkaske: use the `@openstatus/plan` - we have defined here.

apps/server/src/v1/index.ts
```diff
@@ -15,15 +13,16 @@ import { pageSubscribersApi } from "./pageSubscribers";
 import { pagesApi } from "./pages";
 import { statusReportUpdatesApi } from "./statusReportUpdates";
 import { statusReportsApi } from "./statusReports";
+import type { Limits } from "@openstatus/db/src/schema/workspaces/validation";
```
mxkaske: Let's consider the `@openstatus/plan` as source of truth for user plan and limits.

```suggestion
import type { Limits } from "@openstatus/plan";
```

packages/db/src/schema/monitors/monitor.ts
```diff
@@ -11,13 +11,9 @@ import { monitorTagsToMonitors } from "../monitor_tags";
...
-import { workspace } from "../workspaces";
-import {
-  monitorJobTypes,
-  monitorMethods,
-  monitorPeriodicity,
-  monitorStatus,
-} from "./constants";
+import { workspace } from "../workspaces/workspace";
+import { monitorJobTypes, monitorMethods, monitorStatus } from "./constants";
+import { monitorPeriodicity } from "../shared/shared";
```
mxkaske: `/shared/shared` haha

packages/db/src/schema/workspaces/validation.ts
thibaultleouay: but in openstatus plan have dep to openstatus/db 
`import type { WorkspacePlan } from "@openstatus/db/src/schema";` 
so we can't have dep for plan in openstatus/db but i want to validate for insert/select :) 
it's a circular dep it's 😱

packages/db/src/schema/plan/index.ts
mxkaske, line 1: What I don't like is that we have the "free" limits defined here and in `/packages/plan/src/config.ts`. Can we have it centralized? Handling two source of truth is hard.

packages/plans/src/config.ts
```diff
@@ -17,7 +17,8 @@ export const allPlans: Record<
     description: "For personal projects",
     price: 0,
     limits: {
```
mxkaske, line 18: The limits are same to the above defined in `@openstatus/db`.

packages/db/src/schema/workspaces/validation.ts
thibaultleouay, line 1: @mxkaske  here we have the double value

packages/db/src/schema/plan/schema.ts
mxkaske, line 1: Can we keep the utility classes (can be commented) from the versioning? At least to have them here.

packages/db/src/schema/plan/utils.ts
mxkaske, line 1: tbh, I dont think we need those utility functions anymore.

Just use e.g. `workspace.limits["status-pages"]`.

packages/db/src/schema/workspaces/validation.ts
```diff
@@ -3,11 +3,42 @@ import { z } from "zod";
...
 export const workspacePlanSchema = z.enum(workspacePlans);
 export const workspaceRoleSchema = z.enum(workspaceRole);
 
 export const selectWorkspaceSchema = createSelectSchema(workspace).extend({
+  limits: z.string().transform((val) => {
+    const parsed = JSON.parse(val);
+    const result = limitsV1.safeParse(parsed);
+    if (result.error) {
+      // Fallback to default limits
+      return limitsV1.parse({
```
mxkaske, line 18: Please use `allPlans` a limits are fallback and described there
```
      limitsV1.parse(allPlans[workspacePlan].limits)
```

I think we can access the `workspacePlan` to get the fallback limits based on the users plan.

packages/db/src/schema/workspaces/validation.ts
```diff
@@ -3,11 +3,42 @@ import { z } from "zod";
 
 import { workspacePlans, workspaceRole } from "./constants";
 import { workspace } from "./workspace";
+import { limitsV1 } from "../plan/schema";
 
 export const workspacePlanSchema = z.enum(workspacePlans);
 export const workspaceRoleSchema = z.enum(workspaceRole);
 
 export const selectWorkspaceSchema = createSelectSchema(workspace).extend({
+  limits: z.string().transform((val) => {
```
mxkaske, line 13: Are we storing the limits for newer subscribed users? Cuz otherwise they will stay in "free".

packages/api/src/router/stripe/webhook.ts
```diff
@@ -67,6 +68,7 @@ export const webhookRouter = createTRPCRouter({
         subscriptionId: subscription.id,
         endsAt: new Date(subscription.current_period_end * 1000),
         paidUntil: new Date(subscription.current_period_end * 1000),
+        limits: JSON.stringify(getLimits(plan.plan)),
```
thibaultleouay, line 71: @mxkaske  this is where we update the plan

PR #936, opened 2024-07-18
feat: data table + search bar + filter

mxkaske:
https://openstatus-web-git-feat-input-search-open-status.vercel.app/test


mxkaske:
We have created a standalone project for that. Further implementations can be followed in: https://github.com/openstatusHQ/data-table-filters

PR #940, opened 2024-07-18
🐛 CSR baileout 

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/web/src/app/PostHogPageView.tsx
+++ b/apps/web/src/app/PostHogPageView.tsx
@@ -1,29 +1,26 @@
 // app/PostHogPageView.tsx
-'use client'
+"use client";
 
 import { usePathname, useSearchParams } from "next/navigation";
+import { usePostHog } from "posthog-js/react";
 import { useEffect } from "react";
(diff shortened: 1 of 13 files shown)


apps/web/src/app/PostHogPageView.tsx
mxkaske, line 1: why not keeping and changing the logic in `providers/posthog`?

apps/web/src/app/app/layout.tsx
```diff
@@ -1,8 +1,13 @@
 import { SessionProvider } from "next-auth/react";
 
-import { PHProvider, PostHogPageview } from "@/providers/posthog";
+import { PHProvider } from "@/providers/posthog";
+
+const PostHogPageview = dynamic(() => import("@/app/PostHogPageView"), {
```
mxkaske, line 5: you sure about the path?

PR #986, opened 2024-08-20
refactor: removing unwanted files

Ipriyankrajai:
Following the update in #985 where we replaced 'Trigger' with 'Midday,' it seems we overlooked removing the unused files related to 'Trigger.' Let's make sure to clean those up.


Ipriyankrajai:
Hey @thibaultleouay ,

I noticed that we're not using the SVGs from [apps/web/public/assets/partners](https://github.com/openstatusHQ/openstatus/tree/main/apps/web/public/assets/partners). Do we still need these assets, or can they be removed?

thibaultleouay:
🙏

PR #995, opened 2024-09-04
fix: added fix for code assertion

### LINKED ISSUE #988: Status Code assertion field isn't logical

_No response_

### Description

image

1. it should be limited to 599

### Environment

_No response_

nolostra:
## Bug Fix [#988](https://github.com/openstatusHQ/openstatus/issues/988)

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
it basically checks the number if it is above 599 then sets it to 599 else it allows the user to pass

### DIFF
--- a/apps/web/src/components/forms/monitor/section-assertions.tsx
+++ b/apps/web/src/components/forms/monitor/section-assertions.tsx
@@ -176,12 +176,6 @@ export function SectionAssertions({ form }: Props) {
               type="number"
               placeholder="200"
               className="col-span-3"
-              onChange={(e) => {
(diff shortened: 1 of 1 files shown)


mxkaske:
Hey @nolostra, thanks for your contribution! It seems like there is type error. 

Imo, you can get rid of the `onChange` method as you have added the `validate` function. It kinda is a hack to override the `e.target.value`. 

You can you try to run `pnpm tsc` within the `/apps/web` directory to detect any TS error.

CleanShot 2024-09-09 at 13 25 42@2x

PR #996, opened 2024-09-04
🚀 allow every content type

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/checker/ping.go
+++ b/apps/checker/ping.go
@@ -77,7 +77,12 @@ func Ping(ctx context.Context, client *http.Client, inputData request.CheckerReq
 		}
 	}
 	if inputData.Method != http.MethodGet {
-		req.Header.Set("Content-Type", "application/json")
+		head := req.Header
+		_, ok := head["Content-Type"]
+		if !ok {
+			// by default we set the content type to application/json if it's a POST request
+			req.Header.Set("Content-Type", "application/json")
+		}
 	}
 
 	timing := Timing{}
@@ -188,9 +193,16 @@ func SinglePing(ctx context.Context, client *http.Client, inputData request.Ping
 	for key, value := range inputData.Headers {
 		req.Header.Set(key, value)
 	}
+
 	if inputData.Method != http.MethodGet {
-		req.Header.Set("Content-Type", "application/json")
+		head := req.Header
+		_, ok := head["Content-Type"]
+		if !ok {
+			// by default we set the content type to application/json if it's a POST request
+			req.Header.Set("Content-Type", "application/json")
+		}
 	}
+
 	timing := Timing{}
 
 	trace := &httptrace.ClientTrace{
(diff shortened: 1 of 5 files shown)


thibaultleouay:
resolve OPE-360

thibaultleouay:
fixes OPE-360

thibaultleouay:
fixes OPE-360

apps/checker/ping.go
```diff
@@ -77,7 +77,12 @@ func Ping(ctx context.Context, client *http.Client, inputData request.CheckerReq
 		}
 	}
 	if inputData.Method != http.MethodGet {
-		req.Header.Set("Content-Type", "application/json")
+		head := req.Header
+		_, ok := head["Content-Type"]
+		if !ok {
+			// by default we set the content type to application/json if it's a POST request
```
mxkaske, line 83: Should we set by default or allow user to not set any `Content-Type` at all?

apps/web/src/components/forms/monitor/section-requests.tsx
```diff
@@ -40,11 +41,21 @@ interface Props {
...
 
 export function SectionRequests({ form }: Props) {
-  const { fields, append, remove } = useFieldArray({
+  const { fields, append, prepend, remove, update } = useFieldArray({
     name: "headers",
     control: form.control,
   });
+
   const watchMethod = form.watch("method");
+  const [content, setContent] = useState<string>("application/json");
```
mxkaske: Instead of adding a new state, we could use

```
const watchHeaders = form.watch('headers')
const contentType = watchHeaders.find(header => ... === 'Content-Type')
```

packages/ui/src/components/select.tsx
```diff
@@ -5,23 +5,41 @@ import { Check, ChevronDown } from "lucide-react";
...
 
+const selectVariants = cva(
+  "flex w-full items-center justify-between px-3 text-sm focus:outline-none focus:ring-2 focus:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50",
+  {
+    variants: {
+      variant: {
+        default:
+          "border-input h-10 ring-offset-background py-2 placeholder:text-muted-foreground focus:ring-ring rounded-md border bg-transparent ",
+
+        ghost: "focus:ring-accent focus:ring-offset-accent space-x-2",
```
mxkaske: It's a bit messy. E.g. only having a height in 'default' but not in 'ghost' variant.

Personally, I'd just move the classes `border-none text-muted-foreground' to the props of the SelectTrigger and that's it.

apps/checker/ping.go
thibaultleouay: That's for current post request

packages/ui/src/components/select.tsx
thibaultleouay: but i don't want to have so much spacing next the arrow :(

Issue #1012, opened 2024-09-20
Improve server docker build - Add steps 


thibaultleouay:
### Description

https://github.com/openstatusHQ/openstatus/blob/main/apps/server/Dockerfile

In this docker we could use multiple step to speed up build 

first step for dependencies
second step for compilation 
third step to launch 

We are doing something similar for our checker

Dun-sin:
Hi @thibaultleouay can I take this?

PR #1013, opened 2024-09-22
chore: improve server docker build

Dun-sin:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
added 3 steps for dependencies, building and running the code
 

### Related Issue (optional)
#1012

### DIFF
--- a/apps/server/Dockerfile
+++ b/apps/server/Dockerfile
@@ -1,8 +1,7 @@
 FROM oven/bun AS deps
 WORKDIR /app
 
-# Copy only package.json and bun.lockb (if it exists)
-COPY package.json bun.lockb* ./
+COPY . .
 
 RUN bun install --frozen-lockfile --ignore-scripts
 
--- a/apps/server/Dockerfile
+++ b/apps/server/Dockerfile
@@ -1,22 +1,34 @@
-FROM oven/bun
-
+FROM oven/bun AS deps
 WORKDIR /app
+
+# Copy only package.json and bun.lockb (if it exists)
+COPY package.json bun.lockb* ./
+
+RUN bun install --frozen-lockfile --ignore-scripts
+
+FROM oven/bun AS builder
+WORKDIR /app
+
+# Copy dependencies from previous stage
+COPY --from=deps /app/node_modules ./node_modules
 
 # Copy project
 COPY . .
 
 # To keep the image small ;)
-RUN rm -rf /app/apps/docs
-RUN rm -rf /app/apps/screenshot-service
-RUN rm -rf /app/apps/web
-RUN rm -rf /app/packages/api
-RUN rm -rf /app/packages/integrations/vercel
-#  Temp solution for bun
-RUN bun install --ignore-scripts
-WORKDIR /app/apps/server
+RUN rm -rf apps/docs apps/screenshot-service apps/web packages/api packages/integrations/vercel
 
 # Compile the TypeScript application
+WORKDIR /app/apps/server
 RUN bun build --compile --sourcemap src/index.ts --outfile=app
 
+
+FROM oven/bun AS runner
+WORKDIR /app
+
+# Copy the compiled application from the builder stage
+COPY --from=builder /app/apps/server/app ./
+
 EXPOSE 3000
+
 CMD ["./app"]


apps/server/Dockerfile
```diff
@@ -1,22 +1,34 @@
-FROM oven/bun
+FROM oven/bun AS deps
+WORKDIR /app
+
+# Copy only package.json and bun.lockb (if it exists)
+COPY package.json bun.lockb* ./
```
thibaultleouay: We are in a monorepo, don't we need the other package.json from subfolder too ?

apps/server/Dockerfile
```diff
@@ -1,22 +1,34 @@
...
 # To keep the image small ;)
-RUN rm -rf /app/apps/docs
-RUN rm -rf /app/apps/screenshot-service
-RUN rm -rf /app/apps/web
-RUN rm -rf /app/packages/api
-RUN rm -rf /app/packages/integrations/vercel
-#  Temp solution for bun
-RUN bun install --ignore-scripts
-WORKDIR /app/apps/server
+RUN rm -rf apps/docs apps/screenshot-service apps/web packages/api packages/integrations/vercel
```
thibaultleouay, line 18: If we remove them before installing the deps we have a lighter build 
?

Dun-sin: They aren't needed in the build process correct? I may be wrong but I made that assumption when I saw the folder vercel in there

Dun-sin: Also the deps come from package.json, that's why it was installed at the top🤔

I do see what you mean, I would need to separate them, keep those not needed in build at the top and delete the rest later

apps/server/Dockerfile
Dun-sin: Oh my bad, I forgot it's not a universal package.json😅

Dun-sin:
@thibaultleouay looked through the structure again and i don't think we need the other package.json as well

This is just to build the server anyway, apps/web doesn't seem to have a dockerfile i guess because it's running on vercel, others have their own dockerfile

so this dockerfile only applies to the server, and doesn't need any of the other folder doesn't matter the position

thibaultleouay:
@Dun-sin 

If i try to build it locally it can't because the server doest not have the required deps :( 


CleanShot 2024-09-22 at 13 25 14

I would do that to fix the build 

```
FROM oven/bun AS deps
WORKDIR /app

# Copy project
COPY . .

# To keep the image small ;)
RUN rm -rf apps/docs apps/screenshot-service apps/web packages/api packages/integrations/vercel

RUN bun install --frozen-lockfile --ignore-scripts

FROM oven/bun AS builder
WORKDIR /app

# Copy dependencies from previous stage
COPY --from=deps /app/node_modules ./node_modules

COPY . .

# To keep the image small ;)
RUN rm -rf apps/docs apps/screenshot-service apps/web packages/api packages/integrations/vercel

# Compile the TypeScript application
WORKDIR /app/apps/server
RUN bun build --compile --sourcemap src/index.ts --outfile=app


FROM oven/bun AS runner
WORKDIR /app

# Copy the compiled application from the builder stage
COPY --from=builder /app/apps/server/app ./

EXPOSE 3000

CMD ["./app"]

```

Dun-sin:
> @Dun-sin
> 
> If i try to build it locally it can't because the server doest not have the required deps :(
> 
> CleanShot 2024-09-22 at 13 25 14
> 
> I would do that to fix the build
> 
> ```
> FROM oven/bun AS deps
> WORKDIR /app
> 
> # Copy project
> COPY . .
> 
> # To keep the image small ;)
> RUN rm -rf apps/docs apps/screenshot-service apps/web packages/api packages/integrations/vercel
> 
> RUN bun install --frozen-lockfile --ignore-scripts
> 
> FROM oven/bun AS builder
> WORKDIR /app
> 
> # Copy dependencies from previous stage
> COPY --from=deps /app/node_modules ./node_modules
> 
> COPY . .
> 
> # To keep the image small ;)
> RUN rm -rf apps/docs apps/screenshot-service apps/web packages/api packages/integrations/vercel
> 
> # Compile the TypeScript application
> WORKDIR /app/apps/server
> RUN bun build --compile --sourcemap src/index.ts --outfile=app
> 
> 
> FROM oven/bun AS runner
> WORKDIR /app
> 
> # Copy the compiled application from the builder stage
> COPY --from=builder /app/apps/server/app ./
> 
> EXPOSE 3000
> 
> CMD ["./app"]
> ```

you are right, i was hoping i would just copy the necessary, is there a way to set this up using turbo, where the base package.json install every other thing? i think i saw this with typehero as they use turbo as well

thibaultleouay:
Thanks for your work @Dun-sin  but we are going to use dofigen to generate the docker

Dun-sin:
Hi @thibaultleouay, I noticed you closed my PR after merging the other one. It would have been nice to talk through the changes together since I was assigned the issue. I feel like we could've worked on improving my PR rather than just closing it.

Since open source is about learning/community first and then contributing.

Let me know what you think

PR #1046, opened 2024-10-14
🔥 On-demand test

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


apps/server/src/v1/monitors/results/get.ts
```diff
@@ -0,0 +1,120 @@
...
+              latency: z.number().int(), // in ms
+              statusCode: z.number().int().nullable().default(null),
+              monitorId: z.string().default(""),
+              url: z.string().url().optional(),
+              error: z
+                .number()
+                .default(0)
+                .transform((val) => val !== 0),
+              region: z.enum(flyRegions),
+              cronTimestamp: z.number().int().optional(),
```
mxkaske: shouldn't we return the value `timestamp` instead, as it has nothing to do with `cron`.

packages/tinybird/src/os-client.ts
```diff
@@ -460,6 +460,53 @@ export class OSTinybird {
...
+            latency: z.number().int(), // in ms
+            statusCode: z.number().int().nullable().default(null),
+            monitorId: z.string().default(""),
+            url: z.string().url().optional(),
+            error: z
+              .number()
+              .default(0)
+              .transform((val) => val !== 0),
+            region: z.enum(flyRegions),
+            cronTimestamp: z.number().int().optional(),
```
mxkaske: `timestamp`

apps/server/src/v1/monitors/trigger/post.ts
```diff
@@ -0,0 +1,169 @@
...
+          eq(monitor.workspaceId, Number(workspaceId)),
+          isNull(monitor.deletedAt),
+        ),
+      )
+      .get();
+
+    if (!monitorData) {
+      throw new HTTPException(404, { message: "Not Found" });
+    }
+
```
mxkaske, line 83: check if user has exceeded the workspace `limit`

packages/tinybird/src/os-client.ts
thibaultleouay: but we are returning the timestamp is the time at with the request was actually triggered 
crontimestamp is the time when we asked for the request to be trigger
IMO we should return the time we asked for the job to run

apps/server/src/v1/monitors/results/get.ts
thibaultleouay: but the cronTimestamp is the 'scheduled timestamp' 
it's the time we have trigger the job

mxkaske: OK but it shouldn't be `cronTimestamp` as its confusing for the user OR we rename the key. 

Suggestion: let's use `transform` to map cronTimestamp value to timestamp value.

apps/server/src/v1/monitors/trigger/post.ts
```diff
@@ -0,0 +1,185 @@
...
+    const timestamp = Date.now();
+
+    const newRun = await db
+      .insert(monitorRun)
+      .values({
+        monitorId: row.id,
+        workspaceId: row.workspaceId,
+        runnedAt: new Date(timestamp),
+      })
+      .returning();
```
mxkaske, line 109: per month?

apps/server/src/v1/monitors/trigger/post.ts
```diff
@@ -0,0 +1,185 @@
...
+    const { id } = c.req.valid("param");
+    const limits = c.get("limits");
+
+    const count = (
+      await db
+        .select({ count: sql<number>`count(*)` })
+        .from(monitorRun)
+        .where(and(eq(monitorRun.workspaceId, Number(workspaceId))))
+        .all()
+    )[0].count;
```
mxkaske: per month?

And we might wanna add it to the `/v1/checker` api endpoint

thibaultleouay: let's remove api/checker 🤣

apps/server/src/v1/monitors/results/get.ts
thibaultleouay: but let's return the actual time of the request the timestamp make more sense you are right

apps/web/src/config/pricing-table.ts
```diff
@@ -1,10 +1,15 @@
-import type { LimitsV2 } from "@openstatus/db/src/schema/plan/schema";
+import type { LimitsV1, LimitsV2 } from "@openstatus/db/src/schema/plan/schema";
 
 export const pricingTableConfig: Record<
   string,
   {
     label: string;
-    features: { value: keyof LimitsV2; label: string; badge?: string }[];
+    features: {
+      value: keyof LimitsV1 | keyof LimitsV2;
```
mxkaske, line 8: FIXME:

PR #1073, opened 2024-10-29
Update cli endpoint

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/checker/handlers/checker.go
+++ b/apps/checker/handlers/checker.go
@@ -88,6 +88,7 @@ func (h Handler) HTTPCheckerHandler(c *gin.Context) {
 	}
 
 	var called int
+	var result checker.Response
 
 	op := func() error {
 		called++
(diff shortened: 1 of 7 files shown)


apps/checker/handlers/checker.go
```diff
@@ -302,5 +307,5 @@ func (h Handler) HTTPCheckerHandler(c *gin.Context) {
 
 	}
 
-	c.JSON(http.StatusOK, gin.H{"message": "ok"})
```
thibaultleouay, line 305: @mxkaske  we used to return something dummy 
I'm returning the result of the ping instead 
This is useful to get instant result in the /run endoint

apps/checker/handlers/tcp.go
```diff
@@ -196,7 +207,7 @@ func (h Handler) TCPHandler(c *gin.Context) {
 		}
 	}
 
-	c.JSON(http.StatusOK, gin.H{"message": "ok"})
+	c.JSON(http.StatusOK, response)
```
thibaultleouay, line 210: Same as above
Instead of dummy data return the actual result useful for the run endpoint

apps/server/src/v1/monitors/run/post.ts
thibaultleouay, line 1: This is mostly copy pasta from trigger/post might going to refactor it after 🤣

Issue #1139, opened 2024-12-28
[Open for Comments] RSS and Atom Feed Support


washingtonserip:
## Authors
- @mxkaske 
- @thibaultleouay 
- @washingtonserip

# Overview

This document outlines the design and implementation plan for adding RSS 2.0 and Atom 1.0 support to OpenStatus. It will allow you to subscribe to specific public status pages to receive updates on notices and events.

It's open for comments. Development has already started.

## Why RSS 2.0 and Atom 1.0?

OpenStatus currently lacks a standardized method for delivering updates on system statuses. By supporting RSS and Atom, the platform will align with industry standards, ensuring compatibility with a wide range of feed readers and providing users with an efficient way to stay informed.

## Goals

* **Multiple standard support:** Provide support for different feed formats to ensure broad compatibility.  
* **Minute by Minute Updates:** Balanced update frequency for feeds.  
* **Scalability:** Design the system to handle feed from tens of thousands of status pages.  
* **Data Integration:** Source feed data directly from existing SQL database tables.  
* **Filtering:** Allow users to filter feeds by status and kind (incident or maintenance).

## Non-Goals

* Support for formats other than RSS 2.0 and Atom 1.0.  
* Real-time feed updates more frequently than minute by minute.

# Implementation Strategy

## Architecture Layers

* **Data Layer:** SQL database containing incidents and maintenance events.  
* **Business Logic Layer:** Node.js backend to process and generate feeds.  
* **Presentation Layer:** RSS and Atom feed endpoints accessible to users.  
* **Caching Layer:** Redis for efficient feed delivery.

## Dependencies

* **Feed:** [https://www.npmjs.com/package/feed](https://www.npmjs.com/package/feed) 

**Extra:** Replace `rss` with `feed` lib in blog/changelog (and maybe support atom as well?) to remove `rss` from dependencies.

### Incident

* `id`: Unique identifier.  
* `title`: Title of the incident  
* `description`: Summary of the incident.  
* `status`: Current status (active, resolved).  
* `link`: URL to detailed information.  
* `updated_at`: Last updated timestamp.

### Maintenance

* `id`: Unique identifier.  
* `title`: Title of the maintenance.  
* `description`: Summary of the maintenance.  
* `status`: Current status (scheduled, in-progress, completed).  
* `link`: URL to detailed information.  
* `updated_at`: Last updated timestamp.

## APIs

* `GET /feed/rss`: Generate and return RSS 2.0 feed.  
[...]

PR #1149, opened 2025-01-03
feat: rss/atom support on the status pages

washingtonserip:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Description
Implementation of RSS/Atom for the status pages as proposed in [this issue](https://github.com/openstatusHQ/openstatus/issues/1139).

[Example XML files for RSS and Atom feeds.](https://gist.github.com/washingtonserip/757f4c271a861430c35542616090f656)

### Related Issue
https://github.com/openstatusHQ/openstatus/issues/1139

### DIFF
--- a/apps/web/src/app/status-page/[domain]/feed/[type]/route.ts
+++ b/apps/web/src/app/status-page/[domain]/feed/[type]/route.ts
@@ -4,6 +4,8 @@ import { notFound } from "next/navigation";
 import { getBaseUrl } from "../../utils";
 import { statusDict } from "@/data/incidents-dictionary";
 
+export const revalidate = 60;
+
 export async function GET(
   _request: Request,
   props: { params: Promise<{ domain: string; type: string }> }
@@ -47,7 +49,7 @@ export async function GET(
       title: `Maintenance - ${maintenance.title}`,
       link: maintenanceUrl,
       description: maintenance.message,
-      date: maintenance.updatedAt ?? maintenance.createdAt!,
+      date: maintenance.updatedAt ?? maintenance.createdAt ?? new Date(),
     });
   });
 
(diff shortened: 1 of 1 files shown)


apps/web/src/app/status-page/[domain]/feed/[type]/route.ts
Comment (mxkaske) [inline apps/web/src/app/status-page/[domain]/feed/[type]/route.ts:1]: Looks great! Let's potentially add a little cache to it. What do you think?

```
export const revalidate = 60 // 1 minute cache
```

apps/web/src/app/status-page/[domain]/feed/[type]/route.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  page.maintenances.forEach((maintenance) => {
+    const maintenanceUrl = `${baseUrl}/maintenances/${maintenance.id}`;
+    feed.addItem({
+      id: maintenanceUrl,
+      title: `Maintenance - ${maintenance.title}`,
+      link: maintenanceUrl,
+      description: maintenance.message,
+      date: maintenance.updatedAt ?? maintenance.createdAt!,
```
Comment (mxkaske) [inline apps/web/src/app/status-page/[domain]/feed/[type]/route.ts]: Hmm, having `!` will through linting issues and it's unfortunate that `.createdAt` can be `null`. 

Possibly just throw `?? new Date()`?

Comment (washingtonserip) [reply apps/web/src/app/status-page/[domain]/feed/[type]/route.ts]: Good suggestion. Done.

apps/web/src/app/status-page/[domain]/feed/[type]/route.ts
Comment (washingtonserip) [reply apps/web/src/app/status-page/[domain]/feed/[type]/route.ts:1]: Good, we agreed on it and I forgot. Done.

PR #1150, opened 2025-01-03
chore: add react-email templates

mxkaske:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


packages/emails/emails/followup.tsx
```diff
@@ -37,3 +37,4 @@ const FollowUpEmail = () => {
 };
 
 export { FollowUpEmail };
+export default FollowUpEmail;
```
thibaultleouay, line 39: why ?

PR #1158, opened 2025-01-08
feat: add maintenance resource api routes

thomasmol:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

This adds API routes to the server app to expose the maintenance resource:

- `GET` a list of maintenances
- `GET` a single maintenance by id
- `POST` create new maintenance
- `PUT` update an existing maintenance


apps/server/src/routes/v1/maintenance/post.ts
```diff
@@ -0,0 +1,78 @@
...
+      if (input.monitorIds?.length) {
+        await tx
+          .insert(maintenancesToMonitors)
+          .values(
+            input.monitorIds.map((monitorId) => ({
+              maintenanceId: newMaintenance.id,
+              monitorId,
+            })),
+          )
+          .run();
```
mxkaske: We need to validate the `monitorIds` to be within the `workspace.id` scope. Take the [/pages/post.ts](https://github.com/openstatusHQ/openstatus/blob/main/apps/server/src/routes/v1/pages/post.ts#L122) as example.

apps/server/src/routes/v1/maintenance/put.ts
```diff
@@ -0,0 +1,109 @@
...
+        .update(maintenance)
+        .set({
+          ...input,
+          updatedAt: new Date(),
+        })
+        .where(eq(maintenance.id, Number(id)))
+        .returning()
+        .get();
+
+      if (input.monitorIds) {
```
mxkaske: Same as above, please validate the workspace has access to the monitors before creating the relation.

mxkaske (commented):
Hey @thomasmol! Thanks for the PR. 🔥

Just one remark regarding the security for the maintenance to monitor relation.

And it seems there is a [test failing](https://github.com/openstatusHQ/openstatus/actions/runs/12674227203/job/35336784662#step:10:343) but I can have a closer look at the end of the week!

mxkaske:
Btw, same thing regarding `pageId`, we need to validate workspace has access to it.

thomasmol:
Thanks for the feedback @mxkaske ! I added the validation fixes and updated the db seed so the tests for the maintenance should now pass. Not sure about the other failed tests though. Also renamed the dir to `maintenances` to follow the conventions of the other resources.

mxkaske:
Thanks @thomasmol and sorry for the delay!

PR #1160, opened 2025-01-10
🚧 OpsGenie

thibaultleouay:
Only the backend for OpsGenie 

Will need to update for the ui 

We need to store either the region eu or us and the api key


packages/notifications/opsgenie/src/index.ts
```diff
@@ -141,22 +147,25 @@ export const sendTest = async (props: {
...
-      : "https://api.opsgenie.com/v2/alerts";
+  const url = region === "eu"
+    ? "https://api.eu.opsgenie.com/v2/alerts"
+    : "https://api.opsgenie.com/v2/alerts";
+
+  const alert = OpsGeniePayloadAlert.parse({
+    alias: "test-openstatus",
+    message: "Test Alert <OpenStatus>",
+    description:
+      "If can read this, your OpsGenie integration is functioning correctly! Please ignore this alert and delete it.",
```
mxkaske: Typo: `"If YOU can read this,...`

PR #1169, opened 2025-01-23
Automate the dev environment with Devbox

mootoday:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

I separated the work into three commits in case you want to cherry-pick some of the improvements.

- 2b05708342162ab51d5769cdbeffb51d797148a7: Adds `openstatus-dev.db` to the `.gitignore` file
- bde0191aa53937ea01c25985ccd090b597a2d09d: Makes the seed script idempotent
- b97cca0d8be394e38a3d403d2bb66ab5c80f5c1d: Configures Devbox ([github.com/jetpack-io/devbox](https://github.com/jetpack-io/devbox))

To use Devbox, first install it based on [the instructions](https://github.com/jetify-com/devbox?tab=readme-ov-file#installing-devbox) (it's 1 command).

Then run the following command to install all dependencies, start the database, seed it, and start the web app:

```bash
devbox services up
```

### Before this PR

N/A

### After this PR

CleanShot 2025-01-23 at 23 32 24@2x

### Related Issue (optional)

Direct messages with Thibault on Bluesky 😊.

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -3,5 +3,8 @@
   "packages": ["turso-cli@latest", "nodejs@20", "bun@latest", "sqld@latest"],
   "env": {
     "DEVBOX_COREPACK_ENABLED": "true"
+  },
+  "shell": {
+    "init_hook": ["yes | pnpm -v &> /dev/null"]
   }
 }
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.10.6/.schema/devbox.schema.json",
+  "packages": ["turso-cli@latest", "nodejs@20", "bun@latest", "sqld@latest"],
+  "env": {
+    "DEVBOX_COREPACK_ENABLED": "true"
+  },
+  "shell": {
+    "init_hook": [
+      "test -f ./apps/server/.env || cp ./apps/server/.env.example ./apps/server/.env",
+      "test -f ./apps/web/.env || cp ./apps/web/.env.example ./apps/web/.env",
+      "test -f ./packages/db/.env || cp ./packages/db/.env.example ./packages/db/.env"
+    ],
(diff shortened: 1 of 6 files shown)


packages/db/src/seed.mts
```diff
@@ -183,18 +197,23 @@ async function main() {
...
+        monitorId: 1,
+        statusReportId: 2,
+      },
+      {
+        monitorId: 2,
+        statusReportId: 2,
+      },
+    ])
+    .onConflictDoNothing()
+    .run();
```
mootoday, line 216: I added this `.run()` as it was missing previously.

mootoday:
If you want to proceed with the Devbox setup, let me know and I'll update the `README.md` files throughout the repo.

devbox.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.10.6/.schema/devbox.schema.json",
+  "packages": ["turso-cli@latest", "nodejs@20", "bun@latest", "sqld@latest"],
+  "env": {
+    "DEVBOX_COREPACK_ENABLED": "true"
+  },
+  "shell": {
+    "init_hook": [
+      "test -f ./apps/server/.env || cp ./apps/server/.env.example ./apps/server/.env",
```
thibaultleouay: Why do we need this?

mootoday: Might be a leftover from 8 months ago when I first worked on this PR. I can nuke it, let me know.

thibaultleouay: Yeah we should use the pnpm dx task instead of relying on specific devbox task

mootoday: Absolutely agreed. Fixed in ccfd137187d9

thibaultleouay (approved):
🔥

PR #1228, opened 2025-04-10
feat: status report severity

### LINKED ISSUE #683: Add `severity` to status report
Create levels of severity for better evaluation or use a description for the affected services like 'downtime', 'degraded',..

Extend the `StatusReport` with a db column for it.

We might wanna include `maintenance` into the batch.

[Example of severities](https://www.atlassian.com/incident-management/kpis/severity-levels)

mxkaske:
Resolves #683


mxkaske:
We are going to rework that.

PR #1233, opened 2025-04-18
fix: handle missing Google account last name

suraj719:
## Type of change

- [x] 🐛 Bug fix  
- [ ] 🌟 New feature  
- [ ] 🔨 Breaking change  

## Description

This PR fixes a bug where the `lastName` and `name` fields stores `undefined` when a user signs in using Google and does not have a `last name` set on their profile.

The updated logic ensures that:
- `lastName` is an empty string if `params.profile.family_name` is missing.
- `name` only includes the last name if it exists.

This prevents user names like `"John undefined"` from being stored or displayed.

### A picture tells a thousand words (if any)

image

### Before this PR

User name could be saved as `"John undefined"` if Google profile had no last name.

### After this PR

User name is saved as `"John"` in the same scenario.

### Related Issue (optional)

_No related issue linked._

### DIFF
--- a/apps/web/src/lib/auth/index.ts
+++ b/apps/web/src/lib/auth/index.ts
@@ -33,7 +33,9 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
             lastName: params.profile.family_name || "",
             photoUrl: params.profile.picture,
             // keep the name in sync
-            name: ${params.profile.given_name}${params.profile.family_name ? " " + params.profile.family_name : ""},
+            name: `${params.profile.given_name} ${
+              params.profile.family_name || ""
+            }`.trim(),
             updatedAt: new Date(),
           })
           .where(eq(user.id, Number(params.user.id)))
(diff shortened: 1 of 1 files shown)


apps/web/src/lib/auth/index.ts
```diff
@@ -30,10 +30,10 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
           .update(user)
           .set({
             firstName: params.profile.given_name,
-            lastName: params.profile.family_name,
+            lastName: params.profile.family_name || "",
             photoUrl: params.profile.picture,
             // keep the name in sync
-            name: `${params.profile.given_name} ${params.profile.family_name}`,
+            name: ${params.profile.given_name}${params.profile.family_name ? " " + params.profile.family_name : ""},
```
thibaultleouay: Can we use template literal here please :)

mxkaske:
```suggestion
            name: `${params.profile.given_name} ${params.profile.family_name || ""}`.trim(),
```

mxkaske:
Good catch @suraj719 and thanks for your PR!

suraj719:
> Good catch @suraj719 and thanks for your PR!

Thanks @mxkaske ! Just pushed the update🙌

PR #1239, opened 2025-04-27
chore: parallelize follow-up email sending with Promise.all

0xajinkya:
## Type of change

- [ ] 🔨 Breaking change
- [ ] 🐛 Bug fix
- [ ] 🌟 New feature

## Description

Previously, follow-up emails were sent sequentially inside a for...of loop, leading to slower execution. This change uses Promise.all to send emails in parallel, significantly improving the speed of the process.

This PR improves the performance of the sendFollowUpEmails function by replacing the sequential email sending (using a for...of loop) with parallel sending using Promise.all.
This change speeds up the process, especially when there are many users to email, without affecting the functionality.

### Before this PR

Screenshot from 2025-04-28 03-40-11

### After this PR

Screenshot from 2025-04-28 03-40-40

### DIFF
--- a/apps/workflows/src/cron/emails.ts
+++ b/apps/workflows/src/cron/emails.ts
@@ -28,7 +28,7 @@ export async function sendFollowUpEmails() {
     .map((u) => u.email)
     .filter((email) => email !== null)
     //  I don't know why but I can't have both filter at the same time
-    .filter((email) => typeof email === "string" && email.trim() !== "");
+    .filter((email) => email.trim() !== "");
 
   // Chunk emails into batches of 80
   const batchSize = 80;
--- a/apps/workflows/src/cron/emails.ts
+++ b/apps/workflows/src/cron/emails.ts
@@ -28,7 +28,7 @@ export async function sendFollowUpEmails() {
     .map((u) => u.email)
     .filter((email) => email !== null)
     //  I don't know why but I can't have both filter at the same time
-    .filter((email) => email.trim() !== "");
+    .filter((email) => typeof email === "string" && email.trim() !== "");
 
   // Chunk emails into batches of 80
   const batchSize = 80;
--- a/apps/workflows/src/cron/emails.ts
+++ b/apps/workflows/src/cron/emails.ts
@@ -25,19 +25,20 @@ export async function sendFollowUpEmails() {
 
   // Filter valid emails
   const validEmails = users
-    .filter((user) => user.email)
-    .map((user) => user.email);
+    .filter((user): user is { email: string } => typeof user.email === "string" && user.email.trim() !== "")
+    .map(user => user.email);
 
-  // Chunk emails into batches of 100
+  // Chunk emails into batches of 80
   const batchSize = 80;
   for (let i = 0; i < validEmails.length; i += batchSize) {
-    const batch = validEmails.slice(i, i + batchSize) as string[];
+    const batch = validEmails.slice(i, i + batchSize);
     console.log(`Sending batch with ${batch.length} emails...`);
-
     try {
       await email.sendFollowUpBatched({ to: batch });
     } catch {
+      //Stop email send when rate limit error is faced in order to avoid wasteful API calls
       console.error("Rate limit exceeded. Stopping further sends.");
+      break;
     }
   }
 }
--- a/apps/workflows/src/cron/emails.ts
+++ b/apps/workflows/src/cron/emails.ts
@@ -14,16 +14,30 @@ export async function sendFollowUpEmails() {
   date2.setDate(date2.getDate() - 2);
 
   const users = await db
-    .select()
+    .select({
+      email: user.email,
+    })
     .from(user)
     .where(and(gte(user.createdAt, date1), lte(user.createdAt, date2)))
     .all();
 
   console.log(`Found ${users.length} users to send follow ups.`);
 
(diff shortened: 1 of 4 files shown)


thibaultleouay:
Thanks for you PR 

My concern if we parallelize every send email we will hit resend rate limit 
what we could do but it might require a bit more work is to use sendBatch 
https://resend.com/docs/api-reference/emails/send-batch-emails

0xajinkya:
> Thanks for you PR
> 
> My concern if we parallelize every send email we will hit resend rate limit what we could do but it might require a bit more work is to use sendBatch https://resend.com/docs/api-reference/emails/send-batch-emails

Hey @thibaultleouay, I am looking into this. Will keep you updated.

0xajinkya:
Hey @thibaultleouay,

Can you look here?

Following are the key changes:
1. Database optimization: Now only selecting the email field from users instead of fetching complete user records, reducing memory usage and database load.

2. Batch processing: Implemented batching of emails (80 per batch) to prevent overwhelming the email service and improve throughput.

3. Rate limit handling: Added specific detection of rate_limit_exceeded errors, which stops further sending attempts when limits are hit to prevent wasted API calls.

4. Error handling improvements: Enhanced error catching with more specific error types and better logging of failures.

5. Process visibility: Added additional logging to track batch progress.

These changes should make the follow-up email process more reliable and prevent issues when sending to large numbers of users. The batching approach also provides better visibility into the sending process and enables easier recovery when errors occur.

apps/workflows/src/cron/emails.ts
```diff
@@ -14,16 +14,30 @@ export async function sendFollowUpEmails() {
...
-      await email.sendFollowUp({ to: user.email });
+  // Filter valid emails
+  const validEmails = users
+    .filter((user) => user.email)
+    .map((user) => user.email);
+
+  // Chunk emails into batches of 100
+  const batchSize = 80;
+  for (let i = 0; i < validEmails.length; i += batchSize) {
+    const batch = validEmails.slice(i, i + batchSize) as string[];
```
thibaultleouay: not a huge fan of `as string[]`, cant we auto infer it instead?

packages/emails/src/client.tsx
```diff
@@ -40,8 +40,43 @@ export class EmailClient {
...
+          html,
+        }))
+      );
+
+      if (result.error) {
+        if (result.error.name === "rate_limit_exceeded") {
+          throw result.error;
+        }
+
+        console.error(`Batch send error:`, result.error);
```
thibaultleouay: I would also log the error above  move this after line 59

packages/emails/src/client.tsx
```diff
@@ -40,8 +40,43 @@ export class EmailClient {
...
+        return;
+      }
+
+      console.log(`Sent follow up email to ${req.to}`);
+    } catch (err: any) {
```
thibaultleouay: not sure about the try/catch

apps/workflows/src/cron/emails.ts
0xajinkya: Latest change takes care of this. Please have a look.

packages/emails/src/client.tsx
0xajinkya: The thing is, if the batch send fails specifically due to **rate limit**, it is intuitional that subsequent batches will fail as well. So, that's why I am not logging that particular error. And, instead handling it in the loop itself (Logging it and breaking from the loop). Any other error, are being simply consolled.

packages/emails/src/client.tsx
0xajinkya: This is explained in the latest reply to above conversation. Have a look.

0xajinkya:
Any updates @thibaultleouay?

thibaultleouay:
@0xajinkya  Just pushed some update in your pr let me know if you have any question

0xajinkya:
Hey @thibaultleouay, All good, added a minor fix. Also, I am wondering why are we doing this:

` const validEmails = users
    .map((u) => u.email)
    .filter((email) => email !== null)
    //  I don't know why but I can't have both filter at the same time
    .filter((email) => typeof email === "string" && email.trim() !== "");
`

This seems repetitive.

thibaultleouay:
@0xajinkya  We are filtering null first, then removing empty email

apps/workflows/src/cron/emails.ts
```diff
@@ -14,16 +14,33 @@ export async function sendFollowUpEmails() {
...
 
-  for (const user of users) {
-    if (user.email) {
-      await email.sendFollowUp({ to: user.email });
+  // Filter valid emails
+  const validEmails = users
+    .map((u) => u.email)
+    .filter((email) => email !== null)
+    //  I don't know why but I can't have both filter at the same time
+    .filter((email) => typeof email === "string" && email.trim() !== "");
```
thibaultleouay: CleanShot 2025-05-01 at 12 01 49@2x

I dont understand the typeof email here, because the type is already infered ?

0xajinkya: Yes, it was totally not needed. It was error from my end. Changing it.

thibaultleouay:
Thanks for your pr :)

0xajinkya:
It's a pleasure, looking forward more such contributions.

PR #1255, opened 2025-05-19
feat: incident audit logs

mxkaske:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR

### DIFF
--- a/apps/workflows/src/checker/index.ts
+++ b/apps/workflows/src/checker/index.ts
@@ -130,8 +130,20 @@ checkerRoute.post("/updateStatus", async (c) => {
             // incident is already resolved
             break;
           }
-
           console.log(`🤓 recovering incident ${incident.id}`);
+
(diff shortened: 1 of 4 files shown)


apps/workflows/src/checker/index.ts
```diff
@@ -157,7 +164,11 @@ checkerRoute.post("/updateStatus", async (c) => {
           id: `monitor:${monitorId}`,
           action: "monitor.recovered",
           targets: [{ id: monitorId, type: "monitor" }],
-          metadata: { region: region, statusCode: statusCode ?? -1 },
+          metadata: {
+            region: region,
```
thibaultleouay: This could be moved at the top of the switch branch

mxkaske: check again plz

Issue #1258, opened 2025-05-20
Fix dark mode toast close button


thibaultleouay:
### Description

![Image](https://github.com/user-attachments/assets/2b7dc1de-647e-431b-8138-ddae014a829a)

### Environment

```Shell

```

abhiifour:
Please assign :)

kotiyalashwin:
I think this pull request probably solves the issue in all pages because I changed the text color in the main Toaster component which is then used by other pages.

Here: https://github.com/openstatusHQ/openstatus/pull/1259

abhiifour:
@thibaultleouay i think there is a need to update the contributing guidelines.

PR #1259, opened 2025-05-20
Fix: Ensure toast close button is visible in dark mode

kotiyalashwin:
## Type of change

- [ x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Earlier the toaster close button was not visible in dark-mode due to the text-color.
Changed the text-color to black such that it is visible in dark mode also.
Confirmed working in both Light-mode and Dark-mode

### Before this PR
445763428-2b7dc1de-647e-431b-8138-ddae014a829a

### After this PR

https://github.com/user-attachments/assets/38977b0c-1321-4059-bae3-d129ce378a25

### Related Issue (optional)

https://github.com/openstatusHQ/openstatus/issues/1258

### DIFF
--- a/apps/web/src/components/ui/sonner.tsx
+++ b/apps/web/src/components/ui/sonner.tsx
@@ -21,6 +21,8 @@ const Toaster = ({ ...props }: ToasterProps) => {
             "group-[.toast]:bg-primary group-[.toast]:text-primary-foreground",
           cancelButton:
             "group-[.toast]:bg-muted group-[.toast]:text-muted-foreground",
+            closeButton:
+              "group-[.toast]:text-black", 
         },
       }}
       {...props}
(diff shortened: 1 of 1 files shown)


apps/web/src/components/ui/sonner.tsx
```diff
@@ -21,6 +21,8 @@ const Toaster = ({ ...props }: ToasterProps) => {
             "group-[.toast]:bg-primary group-[.toast]:text-primary-foreground",
           cancelButton:
             "group-[.toast]:bg-muted group-[.toast]:text-muted-foreground",
+            closeButton:
+              "group-[.toast]:text-black",
```
mxkaske: can we use shadcn css variables `text-foreground` `text-background` for that?

mxkaske (commented):
Good call! Thanks @kotiyalashwin. Small request to keep it more customizable.

apps/web/src/components/ui/sonner.tsx
kotiyalashwin: All right, I have added 
>text-muted-foreground

for better customizability.

Screenshot from 2025-05-21 17-42-35
Screenshot from 2025-05-21 17-44-07

You can further change the 
```
--muted-foreground
``` 
property as per your need.

Thanks for the feedback.

mxkaske:
Thank you!

PR #1328, opened 2025-08-20
fix: resolve type errors in auth adapter by properly formatting user …

### LINKED ISSUE #812: Fix NextAuth type

_No response_

### Description

We have two ts-expect-error  in adapter.ts 

We should fix the type with next-auth module augmentation 

https://github.com/openstatusHQ/openstatus/blob/45fcc5d6d9cfd59894bdf69d6edb991a776a6557/apps/web/src/lib/auth/adapter.ts#L28

### Environment

_No response_

Nil2000:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Will fix #812

### DIFF
--- a/apps/dashboard/src/lib/auth/adapter.ts
+++ b/apps/dashboard/src/lib/auth/adapter.ts
@@ -13,9 +13,13 @@ import { createUser, getUser } from "./helpers";
 
 export const adapter: Adapter = {
   ...DrizzleAdapter(db, {
+    // @ts-ignore
     usersTable: user,
+    // @ts-ignore
     accountsTable: account,
+    // @ts-ignore
     sessionsTable: session,
+    // @ts-ignore
     verificationTokensTable: verificationToken,
   }),
   createUser: async (data) => {
--- a/apps/dashboard/src/lib/auth/adapter.ts
+++ b/apps/dashboard/src/lib/auth/adapter.ts
@@ -12,7 +12,6 @@ import {
 import { createUser, getUser } from "./helpers";
 
 export const adapter: Adapter = {
-  // @ts-expect-error some issues with types
   ...DrizzleAdapter(db, {
     usersTable: user,
     accountsTable: account,
--- a/apps/dashboard/src/lib/auth/adapter.ts
+++ b/apps/dashboard/src/lib/auth/adapter.ts
@@ -19,12 +19,21 @@ export const adapter: Adapter = {
     sessionsTable: session,
     verificationTokensTable: verificationToken,
   }),
-  // @ts-expect-error some issues with types
   createUser: async (data) => {
-    return await createUser(data);
+    const user = await createUser(data);
+    return {
+      ...user,
+      id: user.id.toString(),
+      email: user.email || "",
+    };
   },
-  // @ts-expect-error some issues with types
   getUser: async (id) => {
-    return await getUser(id);
+    const user = await getUser(id);
+    if (!user) return null;
+    return {
+      ...user,
+      id: user.id.toString(),
+      email: user.email || "",
+    };
   },
 };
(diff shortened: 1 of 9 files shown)


apps/dashboard/src/lib/auth/adapter.ts
```diff
@@ -19,12 +19,21 @@ export const adapter: Adapter = {
...
   createUser: async (data) => {
-    return await createUser(data);
+    const user = await createUser(data);
+    return {
+      ...user,
+      id: user.id.toString(),
+      email: user.email || "",
+    };
   },
-  // @ts-expect-error some issues with types
```
thibaultleouay, line 26: can we also remove this error ? 
https://github.com/openstatusHQ/openstatus/pull/1328/files#diff-438f81713efb2409d37fa605d51cd8a04dfb709b956d0e4a89be1f7b42508332R15

Nil2000: Actually I tried few things like updating same packages related to drizzle-orm,@libsql/client etc but nothing happens. Issue still remained.
Even I tried to reproduce the same error by creating a new turbo repo and include all the version same as the ones used in openstatus but did not trigger any error there.

Nil2000: But the moment I used the schema of user and others (required only for next-auth and drizzle setup), I got errors somthing like these
image

Nil2000: The schema for user are bit different like userIds are used as integer but should be string according to docs as well. Not sure though these are relevant or not for fixing these issues.

Nil2000:
@thibaultleouay  @mxkaske 
Ok currently partial fix is done by making sure all the places (`packages` or `apps`) have same latest version of packages like `@libsql/client`,`drizzle-orm`, `drizzle-kit` etc. and remove wherever not required (like in `apps/screenshot-saver`). Once this is done, while building I only got errors like the following but db one got fixed
image
For these cases,
1. For `usersTable` case, `user` schema need to contain `image` field and `user_id` should be of string type
2. For `accountsTable` and `sessionTable` cases, both table should refer to `user_id` of type string

Nil2000:
Hi @mxkaske ,Would love to know if you have something on your mind for fixing this cause the docker one depends upon this as well(for checking the builds). Also let me know about the suggestion.

thibaultleouay:
@Nil2000  I'm approving build and if it works LFG 🚀

thibaultleouay:
The build is failing @Nil2000  😭
CleanShot 2025-09-01 at 08 56 12@2x

Nil2000:
> The build is failing @Nil2000 😭 CleanShot 2025-09-01 at 08 56 12@2x

Yeah I knew this would happen

To fix these issues, I mentioned few things earlier about some schema changes. Once those done this will go away.
But only if you go for those changes.

thibaultleouay:
We cannot do the schema change no maybe we can add ts ignore ?

Nil2000:
> We cannot do the schema change no maybe we can add ts ignore ?

Ok let me check once locally

Nil2000:
@thibaultleouay Can you try running the build again?

thibaultleouay:
@Nil2000  can you push your latest lockfile please ?

Nil2000:
My bad this should help

thibaultleouay:
I don't get the github action 😭

thibaultleouay:
@Nil2000  we have this error in the action 

```
 ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for '@libsql/client@0.15.9(bufferutil@4.0.8)(utf-8-validate@6.0.5)' in pnpm-lock.yaml
 ``` 
 
 but I don't see any occurence of it 😂

thibaultleouay (approved):
🤬 CI let's merge it

PR #1465, opened 2025-10-20
docker compose to launch openstatus 

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


docker-compose.yaml
```diff
@@ -44,16 +57,40 @@ services:
...
-    #         - 3002:3000
-    #     image: dashboard
-    #     environment:
-    #         - API_URL=http://host.docker.internal:3001
-    #         - NEXT_PUBLIC_OPENPANEL_CLIENT_ID=test
-    #         - OPENPANEL_CLIENT_SECRET=test
-    #         - NODE_ENV=production
-    #     command: .
+    dashboard:
+        build:
```
yipfram, line 61: Is there any way we could do this using a prebuilt image?

This makes it easier for deployment on coolify, cloudron and else

yipfram:
Need help for testing?

docker-compose.yaml
thibaultleouay: Sure, but that's something we will do later. :) Let's make the Docker Compose run with the dev image, then we should use the Next.js build output.

thibaultleouay:
What needs to be done

We need to containerize 
-  apps/workflow
-  app/server 
-  apps/privatelocation 
-  apps/status-page
-  apps/dashboard

When we launch the Docker Compose, all the services should run.

For the self-hosted edition, we only allow private regions (we should update the front end for that, but later).

We need to use the Tinybird Docker image.

In the different project, we should get rid of the hardcoded values. For example, in `/apps/private-location/internal/tinybird/client.go`, we should use the endpoint of our self-hosted Tinybird.

In `apps/workflows/src/cron/index.ts`, this is the place where we launch all the monitoring. We should use crontab or something else; in production, we are using GCP scheduled tasks.

Also, in `apps/workflows/src/cron/checker.ts`, we should bypass the queue and send the request straight to the probe.

Issue #1549, opened 2025-11-13
Upgrade to NextJS 16


aayushpurplesectors:
### Description

Would be cool to see dev perf gains

thibaultleouay:
Do you want to make a pr ?

iamramtin:
@thibaultleouay I believe I have a working version of this properly upgraded to v16, can you assign this issue to me please and I'll open a PR shortly?

iamramtin:
@thibaultleouay Opened PR, lmk if you have any feedback

thibaultleouay:
check the latest build for each 

Build on main for status page,Small improvement 

<img width="2604" height="1960" alt="Image" src="https://github.com/user-attachments/assets/fbfa959f-0493-4cf7-958a-b66e5bde878e" />



On the dashboard not really visible 

<img width="2526" height="2568" alt="Image" src="https://github.com/user-attachments/assets/6e8a62cb-c748-43f1-9ffc-955c7b95f9ec" />

On the landing it's a bit faster: 

<img width="2544" height="2824" alt="Image" src="https://github.com/user-attachments/assets/f70fb9b4-bcd3-48b5-a1c6-77738b289be7" />


Thanks @iamramtin  for the upgrade

PR #1577, opened 2025-11-26
chore: Upgrade to Next.js 16

iamramtin:
## Upgrade to Next.js 16

This PR upgrades OpenStatus from Next.js 15.5.5 to 16.0.4

### Main Changes

**Circular Dependency Handling**
The Drizzle ORM schema has a circular dependency between `incident.ts` ↔ `monitor.ts` (incident references monitor, monitor references incident). This pattern caused issues with Next.js 16's Turbopack.

Initial attempts to "fix" this with Proxy patterns and dynamic `require()` caused runtime errors. To resolve this, Turbopack's scope hoisting optimization in `next.config.ts` had to be disabled:

```typescript
  experimental: {
    turbopackScopeHoisting: false,
  }
```

The Drizzle ORM schema has a circular dependency between incident.ts ↔ monitor.ts. When Turbopack's scope hoisting is enabled, it merges modules together, which breaks circular dependency handling:

  1. Entry module starts and imports both x.js and y.js (merged into one module)
  2. x.js executes and imports a.js
  3. a.js imports something from y.js
  4. But y.js hasn't been marked as executed in the module cache yet (it comes later in the merged module)
  5. This causes the module factory to execute a second time, breaking the circular reference

See https://github.com/vercel/next.js/pull/82827 for related module resolution improvements.

**Breaking API Changes**
  - Updated to Next.js 16's async `cookies()` API
  - Removed deprecated `UnsafeUnwrappedCookies` type
  - Updated all `getPreferredSettings()` callers to handle async signature

**Image Component Updates**
  - Replaced deprecated `layout="fill"` and `objectFit` props
  - Migrated to `fill` attribute and className-based styling (`className="object-contain"`)

**Minor Improvements**
  - Fixed Astro build warnings (added `is:inline` directive, removed unused imports)
  - Added dynamic rendering to `/play/status` page to handle missing TinyBird API keys gracefully
### References
  - [Next.js 16 Release](https://nextjs.org/blog/next-16)
  - [Async Request APIs Breaking Change](https://nextjs.org/docs/app/guides/upgrading/version-16#async-request-apis-breaking-change)
  - [Module Resolution Improvements](https://github.com/vercel/next.js/pull/82827)

https://github.com/openstatusHQ/openstatus/issues/1549


apps/web/src/app/(pages)/(content)/play/status/page.tsx
```diff
@@ -14,6 +14,14 @@ export const metadata: Metadata = {
 export const revalidate = 600;
 
 export default async function PlayPage() {
+  const hasValidApiKey =
+    process.env.TINY_BIRD_API_KEY &&
```
thibaultleouay: can't we use our env package like we are doing it here : 
https://github.com/openstatusHQ/openstatus/blob/40ee67dc9bbbb4d39796e1b7e8b2ae17c61dd77e/apps/web/src/app/api/og/monitor/route.tsx#L11

iamramtin: ah yes, I have updated it

apps/web/src/app/(pages)/(content)/play/status/page.tsx
```diff
@@ -14,6 +15,13 @@ export const metadata: Metadata = {
 export const revalidate = 600;
 
 export default async function PlayPage() {
+  const hasValidApiKey =
```
thibaultleouay, line 18: We need this for compilation?

thibaultleouay:
Can we rename middleware to proxy ?

thibaultleouay:
There's a bug with : 
https://openstatus-status-page-git-fork-iamramtin-ch-14a0a7-open-status.vercel.app/status 

CleanShot 2025-11-26 at 11 06 00@2x
maybe because of lazy loading

apps/web/src/app/(pages)/(content)/play/status/page.tsx
iamramtin: I suppose it's not needed, just thought it would improve the local dev workflow if you don't care about needing tinybird, because the build will fail otherwise every time even with a dummy value:

```
Error: invalid authentication token. Invalid token b'tiny-bird-api-key': 
    at f.fetch (.next/server/chunks/ssr/apps_web_src_lib_tb_ts_b5832e0d._.js:2:862)
    at async Object.getData (.next/server/chunks/ssr/apps_web_src_lib_tb_ts_b5832e0d._.js:2:1251)
    at async g (.next/server/chunks/ssr/[root-of-the-server]__c8b55074._.js:2:524) {
  digest: '974176487'
}
Export encountered an error on /(pages)/(content)/play/status/page: /play/status, exiting the build.
 ⨯ Next.js build worker exited with code: 1 and signal: null
 ```
 
 let me know if you'd prefer I revert this change

thibaultleouay:
I have removed lazyloading let's see if it build on vercel and no there's no issue

iamramtin:
> There's a bug with : [openstatus-status-page-git-fork-iamramtin-ch-14a0a7-open-status.vercel.app/status](https://openstatus-status-page-git-fork-iamramtin-ch-14a0a7-open-status.vercel.app/status)
> 
> CleanShot 2025-11-26 at 11 06 00@2x maybe because of lazy loading

Ah damn I thought I'd resolved the circular dependency issue with the schemas, I'll take a look and see if I can find another solution. I believe this was introduced on next 16 only when they "stabilized" turbopack, this problem did not exists on 15.5.5

iamramtin:
> I have removed lazyloading let's see if it build on vercel and no there's no issue

Did it build successfully without it?

thibaultleouay:
> > I have removed lazyloading let's see if it build on vercel and no there's no issue
> 
> Did it build successfully without it?

nope :( 
CleanShot 2025-11-26 at 11 27 03@2x

PR #1593, opened 2025-12-03
perf(checker): optimize DNS lookup to reduce network calls

shuklavedant01:
## What this PR does

Currently, the `Dns` function calls `lookupA` and `lookupAAAA` separately. Both of these functions trigger `net.LookupIP(host)`, resulting in two identical network requests to the resolver for every check.

## Changes

- Refactored `Dns` to call `net.LookupIP` once.
- Filtered the returned IPs into `A` and `AAAA` slices in memory.
- Removed the redundant `lookupA` and `lookupAAAA` helper functions.

## Benefit

Reduces DNS lookup overhead by ~50% per monitor, improving performance.


thibaultleouay:
@shuklavedant01  thanks for the pr 🤗

PR #1628, opened 2025-12-10
fix dock build

thibaultleouay:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

### Before this PR

### After this PR


iamramtin:
Thanks for adding the missing env vars to the dofigen files!

Heads up, you can also run `docker compose up [service_name]`, so the additional individual docker-compose files won't be necessary

iamramtin (approved):

PR #1646, opened 2025-12-15
 Make entry badges linkable

### LINKED ISSUE #1642: [Improvement] Make entry badges linkable
### Description

everywhere we link relationships as simple badge, we should include links to the corresponding entries (status-page, notifications, monitors, private-locations,...)

Image

aggmoulik:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

Updated the Notifications, Status Reports and Private Locations.

### Before this PR

### After this PR

### DIFF
--- a/apps/dashboard/src/components/data-table/status-reports/columns.tsx
+++ b/apps/dashboard/src/components/data-table/status-reports/columns.tsx
@@ -93,7 +93,7 @@ export const columns: ColumnDef<StatusReport>[] = [
       const value = row.getValue("monitors");
       if (Array.isArray(value) && value.length > 0 && "name" in value[0]) {
         return (
-          <div className="flex cursor-pointer flex-wrap gap-1">
+          <div className="flex flex-wrap gap-1">
             {value.map((m) => (
               <Link href={`/monitors/${m.id}`} key={m.id}>
                 <TableCellBadge value={m.name} />
--- a/apps/dashboard/src/components/data-table/status-reports/columns.tsx
+++ b/apps/dashboard/src/components/data-table/status-reports/columns.tsx
@@ -1,5 +1,8 @@
 "use client";
 
+import type { RouterOutputs } from "@openstatus/api";
+import type { ColumnDef } from "@tanstack/react-table";
+import { ChevronDown, ChevronUp } from "lucide-react";
 import { TableCellBadge } from "@/components/data-table/table-cell-badge";
 import { TableCellDate } from "@/components/data-table/table-cell-date";
 import { TableCellNumber } from "@/components/data-table/table-cell-number";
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/components/data-table/table-cell-badge.tsx
```diff
@@ -38,7 +40,12 @@ export function TableCellBadge({
         {isTruncated ? (
           <Tooltip open={open} onOpenChange={setOpen}>
             <TooltipTrigger
-              onClick={(event) => event.preventDefault()}
+              onClick={(event) => {
```
mxkaske: It's not a good a11y to add `onClick` events to navigate pages. Instead we should use `Link` or `a` components with proper href.

Let's keep this TableCellBadge as is and wrap the Link component everywhere.

aggmoulik:
Hey @mxkaske, I have updated the code using NextJS Link Component but I had to remove the onClick handler from Table Cell Badge as Link will not work due to event.preventDefault

apps/dashboard/src/components/data-table/status-reports/columns.tsx
```diff
@@ -93,9 +93,11 @@ export const columns: ColumnDef<StatusReport>[] = [
       const value = row.getValue("monitors");
       if (Array.isArray(value) && value.length > 0 && "name" in value[0]) {
         return (
-          <div className="flex flex-wrap gap-1">
+          <div className="flex cursor-pointer flex-wrap gap-1">
```
mxkaske: @aggmoulik can we remove the `cursor-pointer` everywhere? If done correctly, anchor tags have the cursor pointer by default.

apps/dashboard/src/components/data-table/table-cell-badge.tsx
```diff
@@ -38,7 +38,6 @@ export function TableCellBadge({
         {isTruncated ? (
           <Tooltip open={open} onOpenChange={setOpen}>
             <TooltipTrigger
-              onClick={(event) => event.preventDefault()}
```
mxkaske, line 41: Does that affect any truncated values? (I wonder why we've added it in the first place)

apps/dashboard/src/components/data-table/status-reports/columns.tsx
aggmoulik: Ahh yes, I haven't checked this. Will fix we need it or not.

apps/dashboard/src/components/data-table/table-cell-badge.tsx
aggmoulik: No, I don't see any issues with it. But yes don't know for sure why we needed to add this as it does nothing.

mxkaske: Yeah 😅

mxkaske:
Thanks @aggmoulik!

PR #1717, opened 2026-01-07
Status Report Updates Not Refreshing in Server Components

### LINKED ISSUE #1676: [UX]: Improvement in Deleting Reports in Overview

_No response_

### Description

**Steps to reproduce:**

https://github.com/user-attachments/assets/34a0f84f-7494-4f93-bb8d-d71afa92fda2

**Expected behavior:**

The row should be removed at real time in the UI table.

**Actual behavior:**

The row is not getting deleted at real time in the UI table.

### Environment

```Shell
OpenStatus live environment.

```

aggmoulik:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

#1676 

### Problem
When creating, updating, or deleting status report updates from client components, the data wouldn't refresh on pages that fetch data server-side (like the Overview page). This was because `queryClient.invalidateQueries()` only works for client-side React Query fetches, not for Next.js Server Components that use `fetchQuery`.

### Solution
Added `router.refresh()` calls to all status report update mutations. This triggers Next.js to re-fetch server components while keeping the existing `invalidateQueries()` for client-side data.

### Changes
- **`data-table.tsx`**: Added `router.refresh()` to the `createStatusReportUpdate` mutation
- **`data-table-row-actions.tsx`**: Added `router.refresh()` to both `updateStatusReportUpdate` and `deleteStatusReportUpdate` mutations

### How It Works Now
- ✅ **Client-side fetched data** (Status Reports page with `useQuery`): `invalidateQueries()` triggers refetch
- ✅ **Server-side fetched data** (Overview page with `fetchQuery`): `router.refresh()` triggers re-render

Both the dedicated Status Reports page and the Overview dashboard now properly update when status report updates are modified.

### Testing
1. Navigate to Overview page (`/overview`)
2. Create/edit/delete a status report update
3. Verify the table updates without requiring a manual page refresh
4. Navigate to Status Reports page (`/status-pages/[id]/status-reports`)
5. Verify updates still work correctly there as well

### Before this PR

### After this PR


apps/dashboard/src/components/data-table/status-report-updates/data-table-row-actions.tsx
```diff
@@ -32,6 +33,8 @@ export function DataTableRowActions({ row }: DataTableRowActionsProps) {
         queryClient.invalidateQueries({
           queryKey: trpc.page.list.queryKey(),
         });
+        // Refresh server components
+        router.refresh();
```
mxkaske: Hey @aggmoulik! 

The idea is goo, but no need to `router.refresh()` to refetch all the queries.

The issue of Status Report Updates not being refreshed happens **only** on the `/overview` page and not on a specific status page because we:

- `refetch` trpc.statusReport.list
- `invalidateQuery` trpc.page.list

Due to the very specific `Date.now()` being used in the trpc method to fetch the status reports on the overview page, it will be impossible to invalidate the query. We will need to rewrite the `startedAt` prop.

See the [overview/page.tsx](https://github.com/openstatusHQ/openstatus/blob/main/apps/dashboard/src/app/(dashboard)/overview/page.tsx#L43)

```ts
  const incidents = await queryClient.fetchQuery(
    trpc.incident.list.queryOptions({
      order: "desc",
      startedAt: {
        gte: new Date(Date.now() - 7 * 24 * 60 * 60 * 1000), // 7 days ago
      },
    }),
  );
```

By default, the `order` is "desc" so we can get rid of that value already. And we can replace the "startedAt" with "period" (e.g. with "7d" enum) to add the `startedAt` logic (`gte: new Date(Date.now() - 7 * 24 * 60 * 60 * 1000)`) in the procedure itself.

That way, we can replace

```ts
router.refresh()
```

with

```ts
queryClient.invalidateQueries({
  queryKey: trpc.pageReport.list.queryKey({ period: "7" }),
});
```

Issue #1732, opened 2026-01-14
Can't install


CoreKit-glitch:
### Description

Hello,

i follow the get started part to install openstatus with docker but i can't rid of pulling images after running

```
docker compose up -d
```

```
 ⠼ Image tinybirdco/tinybird-local:latest       Pulling                                                                  1.5s
 ! Image openstatus/workflows:latest            pull access denied for openstatus/workflows, repository does not exist or may require 'docker login': denied: requested access to the resource is denied 1.4s                                          1.4s
 ! Image openstatus/server:latest               pull access denied for openstatus/server, repository does not exist or may req[+] up 6/35r login': denied: requested access to the resource is denied 1.4s                                             1.4s
 ⠴ Image tinybirdco/tinybird-local:latest [⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀] Pulling                                                    1.6s  ! Image openstatus/workflows:latest pull access denied for openstatus/workflows, repository does not exist or may require 'docker login': denied: requested access to the resource is denied 1.4s r openstatus/dashboard, repository does not exist or may  ! Image openstatus/server:latest pull access denied for openstatus/server, repository does not exist or may require 'docker l[+] up 6/35ed: requested access to the resource is denied 1.4s ied for openstatus/status-page, repository does not exist or ma ⠦ Image tinybirdco/tinybird-local:latest [⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀] Pulling                                                    1.7s  ! Image openstatus/workflows:latest pull access denied for openstatus/workflows, repository does not exist or may require 'docker login': denied: requested access to the resource is denied 1.4s s/dashboard, repository does not exist or may require 'do ! Image openstatus/server:latest pull access denied for openstatus/server, repository does not exist or may require 'docker l[+] up 6/35ed: requested access to the resource is denied 1.4s penstatus/status-page, repository does not exist or may require ⠧ Image tinybirdco/tinybird-local:latest [⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀] Pulling   
```

Images list from docker-compose file are :

        image: ghcr.io/tursodatabase/libsql-server:latest
        image: tinybirdco/tinybird-local:latest
        image: openstatus/workflows:latest
        image: openstatus/server:latest
        image: openstatus/private-location:latest
[...]

PR #1746, opened 2026-01-16
fix(docker): update Dockerfile to install curl for health check

Ariastarcos:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Description

During docker deployment, the healthcheck for the `workflows` service never passes, preventing other services that depend on it from starting. This is caused by `curl` not being in the PATH for the final runtime image (not included in `docker.io/debian:bullseye-slim`).

```
deploy@status:~/docker/openstatus$ docker exec -it openstatus-workflows sh
$ curl -f http://localhost:3000/ping
sh: 1: curl: not found
```

Installing `curl` in the `runtime` stage resolves this.

### Before this PR

### After this PR

### Related Issue (optional)

- Following the install/deployment steps for https://github.com/openstatusHQ/openstatus/issues/1732 (which are the default docker deployment steps), this user will eventually hit this problem as well.

### DIFF
--- a/apps/workflows/Dockerfile
+++ b/apps/workflows/Dockerfile
@@ -116,6 +116,7 @@ COPY \
     --chown=1000:1000 \
     --link \
     "/etc/ssl/certs/ca-certificates.crt" "/etc/ssl/certs/"
+RUN apt update && apt install -y curl && rm -rf /var/lib/apt/lists/*
 USER 1000:1000
 EXPOSE 3000
 ENTRYPOINT ["/app/apps/workflows/app"]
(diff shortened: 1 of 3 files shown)


apps/workflows/Dockerfile
```diff
@@ -116,6 +116,7 @@ COPY \
     --chown=1000:1000 \
     --link \
     "/etc/ssl/certs/ca-certificates.crt" "/etc/ssl/certs/"
+RUN apt update && apt install -y curl && rm -rf /var/lib/apt/lists/*
```
thibaultleouay, line 119: why do you ` rm -rf /var/lib/apt/list/*` ?

Ariastarcos: Because `apt update` downloads package index files and stores them in `/var/lib/apt/lists/` -- these files are only needed during the install step. Once `apt install` finishes, the lists are dead weight in a Docker image.

Issue #1923, opened 2026-03-05
Syntax issue in docker-compose.github-packages.yaml with uppercase characters


TreeSloths:
Due to the issue mentioned here: https://github.com/openstatusHQ/openstatus/issues/1732, I used 
`docker-compose.github-packages.yaml` instead with the command:
`sudo docker compose -f docker-compose.github-packages.yaml --env-file .env.docker up -d`
(You can also just rename it to docker-compose.yaml)

Using that compose file however results in:
```
unable to get image 'ghcr.io/openstatusHQ/openstatus-status-page:latest': Error response from daemon: invalid reference format: repository name (openstatusHQ/openstatus-status-page) must be lowercase
```

The solution is easy, rewrite all instances of `openstatusHQ` to lowercase `openstatushq`
`sed -i -e 's/openstatusHQ/openstatushq/g' docker-compose.github-packages.yaml`

PR #1924, opened 2026-03-05
lowercase GHCR org in docker-compose.github-packages.yaml and README

TreeSloths:
Fixes issue https://github.com/openstatusHQ/openstatus/issues/1923


thibaultleouay:
thanks for the fix @TreeSloths

PR #1925, opened 2026-03-05
Fix workflows curl error by installing it in runtime image

### LINKED ISSUE #1795: Broken curl installation in workflow Dockerfile

_No response_

### Description

Curl for workflow is broken in [this commit](https://github.com/openstatusHQ/openstatus/commit/e17b6a6a7d4ce51957ff44257b36239d7944eec3). Its operation depends on the libcurl library, which is not copied in runtime stage. It is logical to return the package installation as it was done earlier in [the commit](https://github.com/openstatusHQ/openstatus/commit/725dee48a614a6a43de77acc65113b738173f506).

### Environment

```Shell

```

TreeSloths:
Fixes https://github.com/openstatusHQ/openstatus/issues/1795, causing openstatus-workflows image to be broken. 

Curl was copied without its runtime dependency libcurl, causing the health check for openstatus-workflows to constantly fail, resulting in dependent containers to not start, with error messages such as: 
```
dependency failed to start: container openstatus-workflows is unhealthy
```
```
$ sudo docker exec -it openstatus-workflows sh -lc 'curl --version'
curl: error while loading shared libraries: libcurl.so.4: cannot open shared object file: No such file or directory
````

### Changes:

Install ca-certificates + curl in the runtime stage via Dofigen `root`:, so the install runs as root (before `USER 1000:1000`, avoiding permission issues) and pulls in the required shared libraries.
```
fromImage: debian:bullseye-slim
workdir: /app/
root:
  run: apt-get update -q && apt-get install -y -q --no-install-recommends ca-certificates curl && update-ca-certificates && rm -rf /var/lib/apt/lists/*
```
Added `rm -rf /var/lib/apt/lists/*` to keep image size low and some flags for less output. 

### Test

Built apps/workflows/Dockerfile locally and confirmed curl --version works inside the image:
```
$ sudo docker exec -it openstatus-workflows sh -lc 'curl --version'
curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.74.0 OpenSSL/1.1.1w zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.0 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0 librtmp/2.3
Release-Date: 2020-12-09, security patched: 7.74.0-1.3+deb11u16
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets
```

The compose file no longer resulted in errors due to the dependency chain now being healthy.


thibaultleouay:
@claude  can you review this

Issue #1893, opened 2026-02-27
[SELF-HOSTED] No such table "page".


Souipi:
Hi ! I'm trying to setup OpenStatus locally using Docker, but I'm running into database errors with the `status-page` and `dashboard` services. I followed the Docker setup guide: https://github.com/openstatusHQ/openstatus/blob/main/DOCKER.md. However, it seems that the database migrations are not being executed properly.

I ran the following command to verify whether migrations were executed `docker compose logs workflows | grep -A 5 "Running database migrations"` but I get no output, which suggests that migrations did not run.

I then tried to run the migrations manually `docker compose exec workflows sh -c "cd /app/packages/db && bun src/migrate.mts"` but this failed with `sh: 1: bun: not found`.

I have some questions so:

- Are migrations supposed to run automatically in the workflows container?
- Is there an additional step required to trigger database setup?
- Should bun be available inside the container?
- Am  missing an environment variable or configuration step?

Any guidance would be greatly appreciated ^^

Thanks in advance!

## "status-page" service error
```
{
  hostnames: [ 'localhost', '3000' ],
  pathnames: [ '', '' ],
  host: 'localhost:3000',
  urlHost: '0.0.0.0:3000',
  subdomain: '0.0.0.0:3000'
}
{
  pathname: '/',
  type: 'pathname',
  prefix: '0.0.0.0:3000',
  subdomain: '0.0.0.0:3000'
}

Error: Failed query: select "id", "workspace_id", "title", "description", "icon", "slug", "custom_domain", "published", "force_theme", "password", "password_protected", "access_type", "auth_email_domains", "homepage_url", "contact_url", "legacy_page", "configuration", "show_monitor_values", "created_at", "updated_at" from "page" where lower("page"."slug") = ? OR lower("page"."custom_domain") = ?

params: 0.0.0.0:3000,0.0.0.0:3000
    at <unknown> (params: 0.0.0.0:3000,0.0.0.0:3000)
    at w.queryWithCache (.next/server/chunks/[root-of-the-server]__a4301528._.js:462:22018)
    at async w.values (.next/server/chunks/[root-of-the-server]__a4301528._.js:462:27929)
    at async w.get (.next/server/chunks/[root-of-the-server]__a4301528._.js:462:27510)
    at async (.next/server/chunks/[root-of-the-server]__a4301528._.js:581:2777)
    at async n8 (.next/server/chunks/[root-of-the-server]__a4301528._.js:430:59548)
[...]

PR #1934, opened 2026-03-05
Added payment plan change and one-shot container in docs

TreeSloths:
Related to https://github.com/openstatusHQ/openstatus/issues/1893

The one-shot container was used to avoid installing and setting up the necessary tools on the host , keeping as little packages installed as possible on the host server (a standard in my workplace). 

As for the payment plan, this is not something that is revealed in the current documentation, but is of high importance for a self-hosted instance.


thibaultleouay:
that's super cool 🔥

PR #1957, opened 2026-03-09
fix: resolve flaky tests caused by shared state between parallel test files

### LINKED ISSUE #1928: Fix flaky test
Some our tests are failing in ci, 

We should fix it 

https://github.com/openstatusHQ/openstatus/actions/runs/22708645414
https://github.com/openstatusHQ/openstatus/actions/runs/22690520024

johnib:
## Summary

Fixes #1928

The CI test suite has been intermittently failing (~37% failure rate overall, ~10% on main) due to test isolation issues when `bun test` runs test files in parallel against a shared database.

This PR fixes two categories of flaky tests:

### 1. "Return empty" tests using a dedicated workspace

Five `get_all.test.ts` files had "return empty" test cases querying workspace 2 and expecting zero records. Since other tests transiently create/delete records, there's a race condition. The fix moves these tests to workspace 3, which is seeded with no data for these resource types (incidents, maintenances, notifications, pages, status reports).

**Files changed:** `incidents/get_all.test.ts`, `maintenances/get_all.test.ts`, `notifications/get_all.test.ts`, `pages/get_all.test.ts`, `statusReports/get_all.test.ts`

### 2. Validation order in `pages/post.ts`

The page creation route checked the page count limit *before* feature-specific permission checks (password-protection, email-domain-protection). When workspace 2 had transient pages from parallel tests, the count check fired first, returning a generic "Upgrade for more status pages" 402 instead of the specific feature error the test expected. Moving the count check after the feature checks ensures correct error messages regardless of concurrent state.

## Verification

Ran CI **20 consecutive times** on this branch — all 20 passed. This is against a baseline where upstream main fails ~10% of the time on the same tests.

## Test plan

- [x] All existing tests pass (verified 20/20 CI runs)
- [x] No new dependencies
- [x] No behavioral changes in production — only test isolation improvements and a validation reorder that is arguably more correct (check feature permissions before quota)

### DIFF
--- a/apps/server/src/routes/v1/incidents/get_all.test.ts
+++ b/apps/server/src/routes/v1/incidents/get_all.test.ts
@@ -22,7 +22,7 @@ test("return empty incidents", async () => {
   const res = await app.request("/v1/incident", {
     method: "GET",
     headers: {
-      "x-openstatus-key": "2",
+      "x-openstatus-key": "3",
     },
   });
 
--- a/apps/server/src/routes/v1/incidents/get_all.test.ts
+++ b/apps/server/src/routes/v1/incidents/get_all.test.ts
@@ -1,7 +1,57 @@
-import { expect, test } from "bun:test";
+import { afterAll, beforeAll, expect, test } from "bun:test";
+import { db, eq } from "@openstatus/db";
+import { incidentTable, monitor } from "@openstatus/db/src/schema";
 
 import { app } from "@/index";
 import { IncidentSchema } from "./schema";
+
+const TEST_PREFIX = "v1-incident-getall-test";
+let testMonitorId: number;
+let testIncidentId: number;
+
+beforeAll(async () => {
+  await db
+    .delete(incidentTable)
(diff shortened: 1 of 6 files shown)


thibaultleouay:
Instead of using a new workspace ID, it would be better to create some data in a before method.
We have an example here:

https://github.com/openstatusHQ/openstatus/blob/55f2e5326edc30aed6a681618c8e08ec38ff9536/apps/server/src/routes/rpc/services/monitor/__tests__/monitor.test.ts#L46

johnib:
> Instead of using a new workspace ID, it would be better to create some data in a before method. We have an example here:
> 
> https://github.com/openstatusHQ/openstatus/blob/55f2e5326edc30aed6a681618c8e08ec38ff9536/apps/server/src/routes/rpc/services/monitor/__tests__/monitor.test.ts#L46

@thibaultleouay  I updated the PR.

Did some mess apparently in this PR. I used the CI to validate the tests pass over multiple runs, meant to have that on my own fork and then update this PR when all looked good to me.

Apparently it all synced over here.

Anyway - let me know if this change satisfies your request.

PR #1969, opened 2026-03-12
Performance: Fix slow color picker 

aggmoulik:


mxkaske:
nice one! can you check for the `lint` to go through the ci?

PR #1997, opened 2026-03-21
fix: add healthcheck to workflows Dockerfile

### LINKED ISSUE #1795: Broken curl installation in workflow Dockerfile

shubdev19:
Fixes #1795

curl was already installed in the runtime stage but no HEALTHCHECK was defined. Added HEALTHCHECK matching the pattern used in apps/server/Dockerfile.


aggmoulik:
Hey @thibaultleouay Do we need this ?

PR #2006, opened 2026-03-23
refactor: Use `ghcr.io` by default for images in docker compose

Fredkiss3:
## Summary 

This is so that if a user wants to deploy openstatus on their server but doesn't want to build the images, it uses the docker compose images by default and gets a fully working deployment without too much setup.

I've seen that there is a dedicated compose file for that, so maybe this is the expected usage ?

### DIFF
--- a/docker-compose-lightweight.yaml
+++ b/docker-compose-lightweight.yaml
@@ -51,13 +51,12 @@ services:
                 condition: service_healthy
         restart: "no"
 
-
     dashboard:
         container_name: openstatus-dashboard
         build:
             context: .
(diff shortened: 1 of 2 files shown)


thibaultleouay:
Wont it infer it with our build context ? 
or by default it will use the publish one on ghcr ?

Fredkiss3:
> Wont it infer it with our build context ?
> or by default it will use the publish one on ghcr ?

Oh ! you are right... it will not use these images, it will build the images instead of pulling the images from the repo. 

My PR is probably useless then

PR #2026, opened 2026-03-27
fix: scope region metrics to monitor's configured regions

johnib:
## Summary

- The Regions table/chart on the monitor overview page previously passed all possible regions (`monitorRegions`) to `mapRegionMetrics`, which caused unconfigured regions with zero-data rows from Tinybird to appear in the UI
- Replaced `monitorRegions` with `monitor.regions` (plus private locations) so only the monitor's configured regions are shown
- Removed the unnecessary `isLoading` conditional since `mapRegionMetrics` already handles the no-data case by returning placeholder entries

## Test plan

- [ ] Navigate to any monitor's overview page and verify the Regions table only shows regions that the monitor is configured to check
- [ ] Create a monitor with a small subset of regions (e.g., 3) and verify only those 3 appear in the Regions section
- [ ] Verify the loading state still shows the correct placeholder regions while data is being fetched
- [ ] Verify monitors with private locations still show those locations in the Regions table

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -3,40 +3,33 @@ name: Claude Code Review
 on:
   pull_request:
     types: [opened, synchronize, ready_for_review, reopened]
-    # Optional: Only run on specific file changes
-    # paths:
-    #   - "src/**/*.ts"
-    #   - "src/**/*.tsx"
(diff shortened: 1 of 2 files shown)


johnib:
Closing — the fix was lost in the original branch. Reopening with a clean PR.

PR #2028, opened 2026-03-28
fix: resolve biome-ignore lint suppressions

### LINKED ISSUE #826: Fix biome-ignore
### Description

We have several biome lint rule disable 

We should fix them 

```
// biome-ignore
```

johnib:
Hey! This PR tackles the long-standing biome-ignore cleanup from #826.

## What changed

I went through all ~80 `biome-ignore` comments in the codebase and fixed the ones where the underlying code could genuinely be improved:

- **Accumulating spread in reduce** — Replaced the `reduce+spread` pattern with `Object.fromEntries` in the monitor/check route files. Besides satisfying the linter, this is actually faster (O(n) vs O(n²))
- **Node imports** — Added `node:` prefix to `crypto` imports, which is the modern convention
- **Assignments in expressions** — Extracted a couple of inline assignments for readability
- **Untyped test mocks** — Gave proper types to `fetchMock` and a few other `any` uses
- **Dead suppression comments** — Removed `noForEach` ignores that were redundant since the rule is already turned off globally
- **Misc** — Unused variable cleanup, class sorting

About half the comments (~42) are legitimately needed — things like React hook deps that are intentionally omitted, `dangerouslySetInnerHTML` for JSON-LD, and tRPC generics that genuinely need `any`. I left those alone.

Closes #826

### DIFF
--- a/apps/dashboard/src/components/ui/data-table/data-table.tsx
+++ b/apps/dashboard/src/components/ui/data-table/data-table.tsx
@@ -75,8 +75,7 @@ export function DataTable<TData, TValue>({
   pagination,
   setPagination,
 }: DataTableProps<TData, TValue>) {
-  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
-  const [globalFilter, setGlobalFilter] = React.useState<any>();
+  const [globalFilter, setGlobalFilter] = React.useState<unknown>();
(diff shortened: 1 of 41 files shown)


aggmoulik:
We have moved from biome to oxlint. So closing the PR

PR #2030, opened 2026-03-28
fix: scope region metrics to monitor's configured regions

### LINKED ISSUE #725: On monitors page hide region if no data.
### Description

CleanShot 2024-03-22 at 17 47 21
we should only display 2 regions instead of 6

johnib:
Small fix for #725 — the monitors overview page was showing all 40+ possible regions in the Regions table, even if a monitor only checks from 1 or 2.

The root cause was straightforward: after data loads, the code was passing `monitorRegions` (the full list of every possible region) to the metrics display instead of `monitor.regions` (just the ones this monitor is configured to use).

One-variable swap, plus cleanup of the now-unused import and a loading-state conditional that was no longer needed.

Closes #725


mxkaske:
thanks @johnib!

PR #2034, opened 2026-03-30
fix: prevent RPC update handlers from accidentally clearing associations

thibaultleouay:
Proto3 repeated fields default to [] when not sent, making it impossible to distinguish "clear all" from "field not provided". This caused UpdateStatusReport, UpdateMaintenance, and UpdateNotification to silently delete all associations (page components / monitors) when only updating metadata like title or name.

Add optional bool flags (update_page_component_ids, update_monitor_ids) to each update request. Associations are only processed when the flag is explicitly true. Deprecate the direct page_id field on UpdateMaintenanceRequest — pageId is now always derived from components.


thibaultleouay:
@claude  review this

thibaultleouay:
@claude review this again

PR #2086, opened 2026-04-17
refactor(status-page): extract proxy middleware into tested pure stages

mxkaske:
## Summary

- Split the status-page proxy middleware (`src/proxy.ts`) into a pipeline of small pure stages under `src/lib/proxy/`, each with colocated `bun:test` coverage. 137 new tests.
- Every routing and access-control decision is now unit-testable by input/output: locale-mismatch redirect, password gate, email-domain gate, IP-restriction gate, `x-proxy` header rewrite, stpg.dev custom-domain rewrite, default rewrite.
- A `composePageAction` runs stages in priority order and returns a discriminated `Action` (`rewrite` / `redirect` / `passthrough`). `proxy.ts` shrinks to a thin IO shell that dispatches via a single switch.
- Two structured log lines per request (request context + resolved action + `reason`) replace the scattered `console.log`s, improving per-request visibility into which branch fired.
- Behaviour is preserved exactly. Two pre-existing bugs kept with comments for follow-up: the loose `endsWith('/login' | '/restricted')` match, and the post-login redirect's operator-precedence bug that ignores the `?redirect=` param.


mxkaske:
@claude review

PR #2095, opened 2026-04-21
fix(import): mark skipped resources as partial progress

princepal9120:
## Summary

Picks up an import result reporting issue where a phase can be marked as `completed` even when some resources were skipped.

- Ensures phases with a mix of created and skipped resources are reported as `partial`
- Keeps fully successful phases reported as `completed`
- Adds regression coverage for the skipped-resource case

## Fixes applied

- Updated `computePhaseStatus()` in `packages/api/src/service/import.ts`
- Treat skipped resources as partial progress instead of full completion
- Added test coverage in `packages/api/src/service/import.test.ts`

## Why

A phase with skipped resources is not fully complete. Reporting it as `partial` makes the import summary more accurate and easier to trust.

## Test plan

- [x] `pnpm exec biome check packages/api/src/service/import.ts packages/api/src/service/import.test.ts`
- [x] `pnpm exec bun test packages/api/src/service/import.test.ts`

### DIFF
--- a/packages/api/src/service/import.test.ts
+++ b/packages/api/src/service/import.test.ts
@@ -362,6 +362,24 @@ describe("addLimitWarnings", () => {
   });
 });
 
+describe("summary status semantics", () => {
+  test("marks a phase partial when some resources are skipped", () => {
+    const resources = [
+      { sourceId: "1", name: "A", status: "created" as const },
+      { sourceId: "2", name: "B", status: "skipped" as const },
+    ];
+
+    const status = resources.every((r) => r.status === "failed")
+      ? "failed"
+      : resources.some((r) => r.status === "failed") ||
+          resources.some((r) => r.status === "skipped")
+        ? "partial"
+        : "completed";
+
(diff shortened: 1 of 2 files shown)


mxkaske:
@claude review

mxkaske:
thanks @princepal9120!

princepal9120:
Addressed the review feedback in a follow-up commit.

Changes made:
- exported `computePhaseStatus` and tested the real function directly
- added edge-case coverage for:
  - empty resources
  - all created
  - all skipped
  - mixed created/skipped
  - mixed created/failed
  - all failed
- updated `writePagePhase` to use `computePhaseStatus()` so page-phase reporting stays consistent with the rest of the import pipeline
- intentionally treat all-skipped as `completed` to preserve idempotent re-import semantics

Validation:
- `pnpm exec biome check packages/api/src/service/import.ts packages/api/src/service/import.test.ts`
- `pnpm exec bun test packages/api/src/service/import.test.ts`

PR #2099, opened 2026-04-22
fix(subscribers): query active subscriber for subscribe/unsubscribe

mxkaske:
When an email had both an active and an unsubscribed row for the same
page (which can happen legitimately after a resubscribe), two bugs
broke the flow:

- v1 POST /page_subscriber/{id}/update only looked for rows with
  unsubscribedAt IS NOT NULL, so it never detected an active duplicate
  and inserted a second active row. Case differences in the email
  address could also bypass the check and then fail the partial unique
  index.
- RPC UnsubscribeFromPage (by email) fetched the first row matching
  (pageId, email) and updated it, which could target an already
  unsubscribed row and leave the active one untouched, making users
  unable to unsubscribe.

Both queries now filter by channelType = 'email' and
unsubscribedAt IS NULL, matching the partial unique index, and compare
emails case-insensitively. Emails are also stored lowercased on insert.

Adds regression tests for:
- resubscribe after unsubscribe produces exactly one active row
- subscribing while an active row exists returns 409
- case-insensitive duplicate detection
- unsubscribe targets the active row when a prior unsubscribed row
  exists for the same email

https://claude.ai/code/session_0168c5f3tYmM3U4fwspwkTri

### DIFF
--- a/apps/server/src/routes/rpc/services/status-page/__tests__/status-page.test.ts
+++ b/apps/server/src/routes/rpc/services/status-page/__tests__/status-page.test.ts
@@ -1,5 +1,5 @@
 import { afterAll, beforeAll, describe, expect, test } from "bun:test";
-import { db, eq, sql } from "@openstatus/db";
+import { and, db, eq, isNull, sql } from "@openstatus/db";
 import {
   monitor,
   page,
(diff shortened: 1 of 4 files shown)


mxkaske:
@claude review

PR #2101, opened 2026-04-23
feat(services): migrate status-report domain (PR 1/N)

mxkaske:
## Summary

**Stacked on #2100.** First domain migration onto the service layer. Moves every status-report write and read onto `packages/services/src/status-report/` and makes tRPC / Connect / Slack thin adapters. Freezes the `ServiceContext` shape and the single-object `{ ctx, input }` argument convention for every subsequent domain PR.

### Services (`packages/services/src/status-report/`)
- `createStatusReport`, `addStatusReportUpdate`, `resolveStatusReport` (convenience wrapper over add-update with status=resolved), `updateStatusReport` (metadata + associations), `updateStatusReportUpdate` (single-update edit), `deleteStatusReport`, `deleteStatusReportUpdate`, `listStatusReports`, `getStatusReport`, `notifyStatusReport`.
- All mutations wrapped in `withTransaction`, emit `emitAudit` before returning, and throw the typed `ServiceError` hierarchy (NotFound / Forbidden / Conflict / …).
- Zod schemas exported per operation are the contract — `.parse()` runs at every entry as defence-in-depth.
- Internal helpers: `validatePageComponentIds`, `updatePageComponentAssociations`, `getReportInWorkspace`, `getReportUpdateInWorkspace`, `getUpdatesForReport`, `getPageComponentIdsForReport`.
- **Notification is a separate service call.** `notifyStatusReport({ ctx, input })` wraps `dispatchStatusReportUpdate` with workspace-scope + `status-subscribers` plan check, and is invoked as a second awaited call after the mutation — preserving the dashboard's two-call pattern (mandatory under the Edge runtime).

### Surfaces — external contracts preserved

- **tRPC** (`packages/api/src/router/statusReport.ts`): every procedure now calls a service. `create` still returns `{ ...initialUpdate, notifySubscribers }`. `list` and `get` still return the enriched `{ updates, pageComponents, page }` shape the dashboard consumes. `emailRouter.sendStatusReport` becomes a thin wrapper over `notifyStatusReport`.
- **Connect RPC** (`apps/server/src/routes/rpc/services/status-report/`): handler is now proto → parse → service call → convert-back. External Connect API unchanged. Helpers previously exported from this folder (consumed by Slack) are gone; callers go through services.
- **Slack** (`apps/server/src/routes/slack/interactions.ts`): the four status-report branches (create / add-update / update / resolve) go through services. The maintenance branch still writes to db directly — it migrates in PR 2.

### Adapters
- `packages/api/src/service-adapter.ts` — `toServiceCtx(ctx)` + `toTRPCError(err)` (ServiceError → TRPCError with matching codes; ZodError maps to BAD_REQUEST).
- `apps/server/src/routes/rpc/adapter.ts` — `toServiceCtx(rpcCtx)` + `toConnectError(err)`. Synthesises `actor.keyId = "ws:<workspaceId>"` until the auth interceptor captures the real API key id (open question in the plan).
- `apps/server/src/routes/slack/service-adapter.ts` — refetches the full workspace for the pending-action's `workspaceId`, builds a Slack actor context, and renders `ServiceError` → human-readable Slack messages.

### Enforcement
- Biome `noRestrictedImports` scoped to `packages/api/src/router/statusReport.ts` + `apps/server/src/routes/rpc/services/status-report/**`. Bans `@openstatus/db`, `@openstatus/db/src/schema`, `drizzle-orm`, and the three `plan/*` subpaths. Tests are ignored (setup/cleanup legitimately touch db).
- Subpath export `@openstatus/services/status-report`.
- `packages/api` / `apps/server` now depend on `@openstatus/services`.

### Tests
- `packages/services/src/status-report/__tests__/status-report.test.ts` — happy paths, workspace isolation, cascade deletes, cross-workspace `ForbiddenError` / `NotFoundError`, the slack-actor audit branch, and enriched-relations for `list` / `get`.

## Deliberate behaviour changes (small)
- Zod throws on invalid input where the old Connect silently coerced (e.g. out-of-range `limit`). Callers that need clamping can catch.
- `updateStatusReport` with `pageComponentIds: []` now clears associations (same as old Connect's `updatePageComponentIds: true`). Passing `undefined` leaves them untouched.

## Open questions deferred to later PRs
- Connect auth interceptor does not capture API key id today — surfaced as `actor.keyId = "ws:<workspaceId>"`. Fix comes with a small interceptor change in a later PR.
- Slack LLM tool schemas still hand-written; a follow-up can try feeding services' Zod schemas directly through the AI SDK if round-trip is clean.

## Test plan
- [ ] CI runs the new `services` integration tests (seeded turso) alongside existing Connect + tRPC suites
- [ ] Dashboard `statusReport.create` → `emailRouter.sendStatusReport` two-call path still works (no frontend change)
- [ ] Connect `CreateStatusReport` / `AddStatusReportUpdate` preserve `notify` semantics (single RPC call externally, two service calls internally)
- [ ] Slack create / add-update / update / resolve actions (approve / approve_notify / cancel) still round-trip
- [ ] Biome override fires if a new import from `@openstatus/db` sneaks into a scoped file

## Stack

- #2100 — scaffold
- **This PR** — status-report (freezes ctx shape)
- Next: PR 2 maintenance → PR 3 incident → PR 4 monitor → … → final PR (broaden Biome, drop `@openstatus/db` deps, rename `rpc/services/` → `rpc/handlers/`)

### DIFF
--- a/apps/server/Dockerfile
+++ b/apps/server/Dockerfile
@@ -41,6 +41,8 @@ RUN \
     --mount=type=bind,target=packages/assertions/package.json,source=packages/assertions/package.json \
     --mount=type=bind,target=packages/theme-store/package.json,source=packages/theme-store/package.json \
     --mount=type=bind,target=packages/locales/package.json,source=packages/locales/package.json \
+    --mount=type=bind,target=packages/services/package.json,source=packages/services/package.json \
+    --mount=type=bind,target=packages/importers/package.json,source=packages/importers/package.json \
     --mount=type=cache,target=/root/.bun/install/cache,sharing=locked \
     bun install --production --frozen-lockfile --verbose
 
@@ -62,11 +64,11 @@ COPY \
 RUN bun build --compile --sourcemap src/index.ts --outfile=app
 
 # runtime
-FROM debian@sha256:4333240150a6924f878e05ec2c998aec95238010e0e4d2fec6161c90128c4652 AS runtime
+FROM debian@sha256:1a4701c321b1d28b1ff5f0230e766791e4b79b1d4c6c7a70064f4b297b1a330f AS runtime
 LABEL \
     io.dofigen.version="2.7.0" \
     org.opencontainers.image.authors="OpenStatus Team" \
-    org.opencontainers.image.base.digest="sha256:4333240150a6924f878e05ec2c998aec95238010e0e4d2fec6161c90128c4652" \
+    org.opencontainers.image.base.digest="sha256:1a4701c321b1d28b1ff5f0230e766791e4b79b1d4c6c7a70064f4b297b1a330f" \
     org.opencontainers.image.base.name="docker.io/debian:bullseye-slim" \
     org.opencontainers.image.description="REST API server with Hono framework for OpenStatus" \
     org.opencontainers.image.source="https://github.com/openstatusHQ/openstatus" \
--- a/apps/server/dofigen.lock
+++ b/apps/server/dofigen.lock
@@ -15,8 +15,8 @@ effective: |
         path: oven/bun
         digest: sha256:f20d9cf365ab35529384f1717687c739c92e6f39157a35a95ef06f4049a10e4a
       label:
-        org.opencontainers.image.base.name: docker.io/oven/bun:1.3.6
         org.opencontainers.image.stage: install
+        org.opencontainers.image.base.name: docker.io/oven/bun:1.3.6
         org.opencontainers.image.base.digest: sha256:f20d9cf365ab35529384f1717687c739c92e6f39157a35a95ef06f4049a10e4a
       workdir: /app/
       run:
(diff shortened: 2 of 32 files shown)


mxkaske:

Addressed in 539c8219 + also fixes the dashboard build regression that the Vercel preview turned up.

### 1. ✅ `statusReport.list` silent truncation at `limit: 100`

Valid — the old tRPC `list` was unbounded (`findMany` with no limit). My migration silently capped at 100. Fixed:
- `ListStatusReportsInput.limit` now `min(1)` with no max at the schema level
- tRPC passes `limit: 10_000` (sentinel "effectively unlimited" for dashboard use)
- Connect handler keeps its own `max=100` cap externally — external API behaviour unchanged

### 2. ✅ `LIMIT_EXCEEDED` → `Code.ResourceExhausted`

Valid — `PermissionDenied` means "authenticated but not entitled"; quota is `ResourceExhausted` semantically per the Connect/gRPC code table. Fixed.

### 3. ❌ Biome `include` key — no change

This one I'm pushing back on. In Biome 1.8.3 (the version pinned here), `OverridePattern.include` is the documented glob key — the schema entry reads:

> "A list of Unix shell style patterns. The formatter will include files/folders that will match these patterns."

The override *is* firing — `pnpm biome check` flagged the Connect test file under `apps/server/src/routes/rpc/services/status-report/__tests__/` the moment the `noRestrictedImports` rule landed (which is why the rule now has the `__tests__` ignore). `includes` (plural) is the Biome 2.x spelling; switching would be a false migration here.

If Biome 2 lands in this repo later, this key changes as part of that upgrade — worth a `biome migrate` run at that point.

### 4. 🐛 Dashboard build regression (bonus — not from the Cubic review)

Vercel's preview build caught a TS error I missed: `row.original.page.customDomain` in `components/data-table/status-reports/data-table-row-actions.tsx:35`. My service migrated the return type from the old zod-validated `page: Page` (non-null) to `page: Page | null` (accurate to the nullable `pageId` FK). The dashboard narrows on `pageId` earlier in the component, but TS can't cross-correlate `pageId` nullability with `page` nullability.

Fixed by narrowing `page` to non-null in the tRPC `get` / `list` wrappers with a small `narrowPage` helper that throws on the shouldn't-happen case (status report without its page) — matches the old zod-parse behaviour exactly.

mxkaske:
@claude review

mxkaske:
@claude review and check again

PR #2103, opened 2026-04-23
feat(services): migrate maintenance domain (PR 2/N)

mxkaske:
## Summary

**Stacked on #2101.** Second domain onto the service layer. Every maintenance write and read goes through `@openstatus/services/maintenance`; tRPC / Connect / Slack are thin adapters. Same shape as the status-report PR.

### Services (`packages/services/src/maintenance/`)
- `createMaintenance`, `updateMaintenance`, `deleteMaintenance`, `listMaintenances`, `getMaintenance`, `notifyMaintenance`.
- Date range invariant (`from < to`) enforced via Zod refine on `CreateMaintenanceInput` and again inside `updateMaintenance` when the partial update could cross the invariant.
- All mutations wrapped in `withTransaction`, emit `emitAudit` before returning.
- Internal helpers (`validatePageComponentIds`, `updatePageComponentAssociations`, `getMaintenanceInWorkspace`, `getPageComponentIdsForMaintenance`, `getPageComponentsForMaintenance`) duplicated from status-report for this PR. Noted as the extraction trigger once a third consumer arrives.

### Surfaces — external contracts preserved
- **tRPC** (`packages/api/src/router/maintenance.ts`): `delete` / `list` / `new` / `update` call services. `new` still returns `{ ...maintenance, notifySubscribers }` for dashboard consumers. `list` passes `limit: 10_000` (same sentinel strategy as the status-report list after the PR 1 Cubic fix).
- **Connect RPC** (`apps/server/src/routes/rpc/services/maintenance/`): proto → parse → service → convert; external Connect API unchanged.
- **Slack** (`interactions.ts`): `createMaintenance` branch goes through services. Also extracts `getPageUrl` / `getReportUrl` into `apps/server/src/routes/slack/page-urls.ts` so `interactions.ts` becomes db-free and can be added to the Biome scope.
- **emailRouter** (`packages/api/src/router/email/index.ts`): `sendMaintenance` becomes a thin wrapper over `notifyMaintenance`.

### Enforcement
- Biome `noRestrictedImports` scope now also covers:
  - `packages/api/src/router/maintenance.ts`
  - `apps/server/src/routes/rpc/services/maintenance/**`
  - `apps/server/src/routes/slack/interactions.ts`

  Carve-outs (outside scope): `slack/page-urls.ts`, `slack/service-adapter.ts`, `slack/confirmation-store.ts`, `slack/workspace-resolver.ts` — all legitimately need db for transport/URL/auth concerns.
[...]

PR #2104, opened 2026-04-23
feat(services): migrate incident domain (PR 3/N)

mxkaske:
## Summary

**Stacked on #2103.** Third domain onto the service layer. tRPC-only — no Connect handler, no Slack path — which makes this the first PR that exercises the `ServiceContext` actor variants without adapter-specific surface code.

### Services (`packages/services/src/incident/`)
- `acknowledgeIncident`, `resolveIncident`, `deleteIncident`, `listIncidents`, `getIncident`.
- `acknowledge` / `resolve` stamp the acting user's id onto `acknowledged_by` / `resolved_by` via the new `tryGetActorUserId(actor)` helper on `ServiceContext`. Non-user actors (system, webhook, API keys without a linked user) stamp `null`.
- Already-acknowledged / already-resolved → `ConflictError`. Matches the existing tRPC's `BAD_REQUEST` semantics (maps back to 400).
- All mutations wrapped in `withTransaction`, emit `emitAudit` before returning.

### list.ts — already-good pattern
- Batch-enriches monitors via a single IN query against distinct `monitorId`s. No N+1 even with the `10_000` sentinel tRPC passes.

### Surface (tRPC only)
- `packages/api/src/router/incident.ts`: all four procs call services.
  - `delete` catches `NotFoundError` to preserve the pre-migration idempotent behaviour.
  - `list` narrows `monitor` to non-null in the return type — every incident has an associated monitor via FK; a null here is a data-integrity signal.
  - `acknowledge` / `resolve` still return `true` to match the old contract.

### Context helper
- `tryGetActorUserId(actor)` added to `packages/services/src/context.ts` + re-exported from the root barrel. Used now for `acknowledged_by` / `resolved_by`; will generalise to other stamp columns as more domains migrate.

### Enforcement
- Biome `noRestrictedImports` scope adds `packages/api/src/router/incident.ts`.
- Subpath export `@openstatus/services/incident`.

### Tests
- `__tests__/incident.test.ts` — happy paths for acknowledge / resolve / delete, `ConflictError` on double-acknowledge / double-resolve, workspace isolation across all mutations, list workspace isolation, and the batched monitor enrichment with two incidents sharing one monitor.

## Test plan
- [ ] Dashboard `/monitors/:id/incidents` and `/overview` incident lists
- [ ] `acknowledge` / `resolve` / `delete` mutations from the incidents data-table still work

## Stack

- #2100 scaffold → #2101 status-report → #2103 maintenance → **this PR** incident
- Next: PR 4 monitor (largest; may split 4a HTTP / 4b TCP+DNS)


mxkaske:

Both valid — fixed in `ea59c630`.

### ✅ 1. TOCTOU on `acknowledge` / `resolve`

The pattern was read-then-update:
```ts
const existing = await getIncidentInWorkspace(...);
if (existing.resolvedAt) throw ConflictError;
await tx.update(incidents).set({...}).where(eq(id, …));  // unconditional
```

Two concurrent resolvers both pass the pre-check and both update — last write wins. Replaced with a conditional update:
```ts
await tx.update(incidents)
  .set({ resolvedAt: now, … })
  .where(and(eq(id, …), isNull(resolvedAt)))   // atomic CAS
  .returning()
  .get();
if (!updated) throw new ConflictError("Incident already resolved.");
```

The loser of the race gets no row back and surfaces the same `ConflictError` the pre-read would have raised. Same fix in `acknowledge.ts`. Dropped the now-unreachable `InternalServiceError` branch in both.

### ✅ 2. Monitor enrichment missing workspace filter

Valid — `incident.monitorId` has no FK against workspace ownership, so a cross-workspace pointer (however unlikely in current data) could leak the other workspace's monitor row. `enrichIncidentsBatch` now takes a `workspaceId` arg and scopes the IN query with `eq(monitor.workspaceId, workspaceId)`. Defence-in-depth.

Note: the pre-migration tRPC had the same gap (drizzle's `with: { monitor: true }` had no workspace filter). The fix is strictly tightening, not preserving old behaviour.

mxkaske:
@claude review

PR #2105, opened 2026-04-23
feat(services): migrate monitor domain tRPC (PR 4/N)

mxkaske:
## Summary

**Stacked on #2104.** Migrates the **tRPC** monitor router — all 14 procedures — onto `@openstatus/services/monitor`. The Connect RPC monitor handler and the v1 REST `monitors/*` endpoints stay untouched for now; they're separate external-API surfaces with their own test suites and warrant their own follow-up PRs.

### Services (`packages/services/src/monitor/`)

CRUD + relation management + list / get:
- `createMonitor`, `cloneMonitor`, `deleteMonitor`, `deleteMonitors` (bulk soft-delete), `getMonitor`, `listMonitors`.
- `updateMonitorGeneral` — keeps the existing tRPC jobType-agnostic update (HTTP ↔ TCP ↔ DNS switch allowed, called out as a code smell in the explore but kept for dashboard-compatibility).
- Field-specific setters: `updateMonitorRetry`, `updateMonitorFollowRedirects`, `updateMonitorOtel`, `updateMonitorPublic`, `updateMonitorResponseTime`.
- Relation setters (transactional + validated): `updateMonitorSchedulingRegions` (plan-limit checks + private-location validation), `updateMonitorTags`, `updateMonitorNotifiers`.
- `bulkUpdateMonitors` — batched toggle of `public` / `active`.
- All mutations inside `withTransaction`, emit `emitAudit`. Soft delete cascades to `monitor_tag_to_monitor`, `notifications_to_monitors`, and `page_component` rows.

### Shared internals
- `validateTagIds` / `validateNotificationIds` / `validatePrivateLocationIds` — workspace-scoped, dedupe input.
- `pickDefaultRegions(workspace)` — ports the old "random 4 free / 6 paid, excluding deprecated" logic.
- `serialiseAssertions` / `headersToDbJson` — moved out of the router.
- `countMonitorsInWorkspace` for quota checks in `create` / `clone`.

### list.ts — batched enrichment
`listMonitors` runs two IN queries (tags + incidents) regardless of row count. `getMonitor` reuses the same batch path with an additional pair of IN queries for `notifications` / `privateLocations`. Pairs well with the `10_000` sentinel tRPC passes.

### Surface — external contract preserved
`packages/api/src/router/monitor.ts` — every procedure is a thin wrapper:
- `delete` catches `NotFoundError` for idempotency (same pattern as status-report / maintenance / incident).
- `new` and `updateGeneral` keep the `testHttp` / `testTcp` / `testDns` pre-save endpoint check at the tRPC layer. Services unconditionally save; pre-validation is caller-specific UX.

## Deliberately out of scope

- **Connect RPC monitor handler** (`apps/server/src/routes/rpc/services/monitor/`) — still uses its own `validators.ts` / `limits.ts` / converters. Migrating it to services is a self-contained follow-up (`4b`), doesn't block this PR.
- **v1 REST** (`apps/server/src/routes/v1/monitors/*`) — four separate endpoints that insert monitors directly. Another external surface; dedicated PR.
- **`packages/api/src/service/import.ts` monitor writes** — covered by the plan's dedicated PR 9.

## Enforcement

- Biome `noRestrictedImports` scope adds `packages/api/src/router/monitor.ts`. The Connect monitor handler stays out of scope until 4b.
- Subpath export `@openstatus/services/monitor`.

## Tests

- `__tests__/monitor.test.ts` — create (http / tcp / dns), delete cascade, bulk delete, clone, tags / notifiers validation (forbidden cross-workspace, dedupe), list / get workspace isolation + soft-delete hiding, `updateMonitorGeneral` round-trip.

## Test plan
- [ ] Service integration tests
- [ ] Dashboard `/monitors` list + get + create flow
- [ ] Dashboard edit forms — tags / notifiers / regions / retry / public / otel / response-time
- [ ] Dashboard clone + delete (single + bulk)
- [ ] Connect RPC monitor suite unaffected (unchanged)

## Stack

- #2100 scaffold → #2101 status-report → #2103 maintenance → #2104 incident → **this PR** monitor (tRPC)
- Follow-ups: Connect monitor handler migration (4b), then PR 5 notification → PR 6 page-component → …


mxkaske:

All five valid — fixed in `1d7414b1`.

### ✅ 1. `timeout` / `degradedAfter` bounds (P1)

My `UpdateMonitorResponseTimeInput` had `z.number().int()` with no bounds, while the canonical `insertMonitorSchema` caps both fields at `0..60_000` ms. Aligned to match — invalid values are rejected at the service boundary instead of being persisted.

### ✅ 2. `jsonBody` assertions silently dropped (P2)

`JsonBodyAssertion` exists in `@openstatus/assertions` (line 348 of `v1.ts`); I just didn't map it in `serialiseAssertions`. The old tRPC router had the same gap, so this is a strict improvement — adding the missing branch means `jsonBody` assertions now round-trip correctly to the stored JSON.

### ✅ 3. Clone inherits stale `status` (P2)

`cloneMonitor` was spreading `{ ...source }` into the insert, carrying the source's current `status` (which could be `error` / `degraded`). Destructured `status` out and forced `"active"` — freshly cloned monitors settle on their own first check. The old tRPC had the same bug; this tightens.

### ✅ 4. Single `delete` re-runs side effects on soft-deleted rows (P2)

`deleteMonitor`'s pre-check was missing `isNull(monitor.deletedAt)`, so a repeated delete would re-run the tag / notifier / page-component cascades and emit a duplicate audit. Added the filter; the second delete now returns `NotFoundError`, which the tRPC wrapper swallows → external behaviour is still "idempotent" from the caller's perspective, but without the duplicate side effects internally.

### ✅ 5. List/get enrichment missing workspace scope (P2)

Threaded `workspaceId` through `enrichMonitorsBatch` and added `eq(…workspaceId, workspaceId)` to **all four** relation queries — tags, incidents, notifications, private-locations. Defence-in-depth: none of those tables enforce workspace ownership at the FK level, so inconsistent data wouldn't leak. Same pattern as the monitor-enrichment fix on #2104.

---

Local typecheck + biome clean on all five changes.

mxkaske:
@claude review

mxkaske:
@claude review

PR #2106, opened 2026-04-23
feat(services): migrate notification CRUD (PR 5/N)

mxkaske:
## Summary

**Stacked on #2105.** Fifth domain onto the service layer. Migrates the four CRUD procedures of the notification tRPC router — `list`, `new`, `updateNotifier`, `delete` — to `@openstatus/services/notification`. The three integration-helper procedures (`sendTest`, `createTelegramToken`, `getTelegramUpdates`) stay inline at the tRPC layer and are explicitly out of scope for this PR.

### Services (`packages/services/src/notification/`)
- `createNotification` — three-layer plan gate: (1) channel count ≤ `notification-channels`, (2) plan flag enabled for gated providers (`sms` / `pagerduty` / `opsgenie` / `grafana-oncall` / `whatsapp`), (3) `NotificationDataSchema` validation of the channel payload. Also validates monitor ids are in-workspace + not soft-deleted.
- `updateNotification` — same validation rules as create; replaces `name` / `data` / monitor associations in a single transaction.
- `deleteNotification` — hard-delete. FK cascade takes care of `notifications_to_monitors`. The tRPC wrapper catches `NotFoundError` to preserve the pre-migration idempotent behaviour.
- `listNotifications` / `getNotification` — batched IN query enriches monitors per notification; monitor enrichment is workspace-scoped + filters soft-deleted monitors (applying the Cubic defence-in-depth lesson from PR 3/4 preemptively).
- All mutations inside `withTransaction`, emit `emitAudit`.

### Surface migrated
`packages/api/src/router/notification.ts` — the four CRUD procedures become thin service wrappers.

## Deliberately out of scope

- **`sendTest` migration** — the dispatch switch imports from 10 `@openstatus/notification-*` packages. Moving it into services means pulling those as direct deps; the plan's phrasing ("Channel CRUD + test-dispatch") leaves space to defer. Follow-up PR.
- **`createTelegramToken` / `getTelegramUpdates`** — redis + external Telegram API helpers. Transport UX, not domain operations; keeping them at the router stays aligned with the plan's carve-outs (like `telegram-updates.ts`).
- **Biome scope for `notification.ts`** — deferred. The file still imports `@openstatus/db/src/schema` for the `sendTest` provider data schemas, so turning the ban on now would flag those. The scope will land with the sendTest migration.
- **Connect RPC notification handler** — stays on its own helpers. Follow-up matching the Connect monitor deferral from PR 4.

## Tests

`__tests__/notification.test.ts`:
- `createNotification`: happy path + audit, `ValidationError` on malformed data, `LimitExceededError` when the free plan blocks `pagerduty`, `ForbiddenError` on cross-workspace monitor attachment.
- `updateNotification`: association replacement (`monitors: []` clears), cross-workspace `NotFoundError`.
- `deleteNotification`: row removed.
- `list` / `get`: enriched monitors visible to owner, invisible cross-workspace.

## Test plan
- [ ] Dashboard `/notifications` list + create + edit + delete flows
- [ ] `trpc.notification.sendTest` for each provider type still works (unchanged)

## Stack

- #2100 scaffold → #2101 status-report → #2103 maintenance → #2104 incident → #2105 monitor (tRPC) → **this PR** notification (CRUD)


mxkaske:

Both valid — fixed in `565e9b5a`.

### ✅ 1. Update flow missing plan-gated provider validation

`createNotification` was calling `assertProviderAllowed`; `updateNotification` wasn't. A user who created a pagerduty channel on the team plan and then downgraded to free could still edit the config — the plan gate only fired on new creates. Added `assertProviderAllowed(ctx.workspace, existing.provider)` to the update flow (uses the stored provider since the API doesn't allow provider changes on update).

Matches the intended semantics: if your plan doesn't entitle you to the provider, you can't tweak it. The old tRPC had the same gap; this is a strict tightening.

### ✅ 2. `data` validated independently of `provider`

`NotificationDataSchema` is a `z.union([...])`, so `{ provider: "discord", data: { slack: "…" } }` passed the top-level parse because the `slack` variant matched — even though the payload key doesn't match the declared provider. Input like this would round-trip to a record with mismatched metadata.

`validateNotificationData` now takes the provider as its first arg and asserts `provider in data` after the union parse. Every canonical data schema in this package uses the provider name as its required key (`discord`, `slack`, `pagerduty`, `google-chat`, etc.), so this single check is exhaustive. Applied in both `create` (with the input's `provider`) and `update` (with the stored `existing.provider`, since the update API doesn't support provider changes).

Added a test for the mismatch case: `provider: "discord"` with `data: { slack: "…" }` now rejects with `ValidationError` instead of silently passing.

---

Local typecheck + biome clean.

mxkaske:
## Follow-up on the last fix

Fair catch — fixed in \`141be110\`.

My previous attempt did `NotificationDataSchema.safeParse(data)` + `provider in data`. That misses the shape check on the provider's own payload: `{ discord: "not-a-url", slack: "valid-url" }` parses against the union (the slack variant matches and silently ignores the extra `discord` key), then my key check sees `"discord"` and lets it through.

Replaced with a provider-specific schema lookup:
```ts
const providerDataSchemas = {
  discord: discordDataSchema,
  slack: slackDataSchema,
  pagerduty: pagerdutyDataSchema,
  // …all 12
} satisfies Record<NotificationProvider, ZodType>;

const parsed = providerDataSchemas[provider].safeParse(data);
```

Each canonical schema is keyed by its provider name and validates the shape + content of the value — one parse covers both the mismatched-provider case and the malformed-payload case. The `satisfies Record<NotificationProvider, ZodType>` clause keeps the map exhaustive at the type level; adding a new provider enum member becomes a compile error until the map gets its entry.

Added the exact case from the finding as a test — invalid `discord` URL alongside a valid `slack` URL now rejects with `ValidationError` instead of slipping through.

mxkaske:
@claude review

PR #2108, opened 2026-04-23
feat(services): Connect RPC notification handler catch-up

mxkaske:
## Summary

**Stacked on #2107.** Catch-up PR — responding to the observation that my PRs from #2105 onwards had been narrowing scope to tRPC only, with Connect RPC / Slack handlers implicitly deferred. This closes the gap for **notification**.

Audit of the stack before this PR:

| Domain | tRPC | Connect | Slack |
|---|---|---|---|
| status-report | ✅ | ✅ | ✅ |
| maintenance | ✅ | ✅ | ✅ |
| incident | ✅ | (no Connect) | (no Slack) |
| monitor | ✅ | ❌ | (no Slack) |
| notification | ✅ | ❌ | (no Slack) |
| page-component | ✅ | (no Connect) | (no Slack) |

Slack is done (only touches status-report + maintenance). Connect monitor is big — ~880 lines with 6 jobType-specific methods; parked as dedicated follow-up. This PR does Connect notification.

## What lands

`apps/server/src/routes/rpc/services/notification/index.ts` — the five CRUD methods now delegate to `@openstatus/services/notification`:

- `createNotification` → service (plan-count limit + provider plan gate + data-schema validation happen in the service itself; the Connect-side `checkNotificationLimit` / `checkProviderAllowed` / `validateProviderDataConsistency` calls are gone from this path).
- `getNotification`, `listNotifications`, `deleteNotification` — proto → service → proto.
- `updateNotification` — Connect's update is partial; the service expects a full payload, so the handler reads the existing record via the service and fills in the omitted fields (name / data / monitors) before applying.

**Left inline at Connect**:
- `sendTestNotification` (external HTTP via `test-providers.ts`).
- `checkNotificationLimit` RPC method (returns count info via `./limits.ts`; pure query, not a mutation).

File shrinks from 402 → 193 lines (-344 / +132 net).

## Biome

Added `apps/server/src/routes/rpc/services/notification/index.ts` to the scope. A directory glob doesn't fit yet because `limits.ts` and `test-providers.ts` still need direct db access — closed when those migrate.

## Deferred

- **Connect monitor handler** (~880 lines, HTTP/TCP/DNS create/update pairs + trigger/status/summary integration methods) — requires a much bigger refactor. Dedicated PR 4b, tracked separately.

## Test plan
- [ ] Existing Connect `notification.test.ts` still green (external contract unchanged)
- [ ] Dashboard notification flows unchanged via tRPC

## Stack

[...]

PR #2109, opened 2026-04-23
feat(services): migrate page authoring (PR 7/N)

mxkaske:
## Summary

**Stacked on #2108.** Seventh domain. Migrates the 13 authoring procedures of `pageRouter` **and the 5 page-CRUD methods of the Connect `status-page` handler** onto `@openstatus/services/page`. The public viewer domain (`statusPage.ts`) and the remaining 13 Connect methods (components / groups / subscribers / viewer) still warrant dedicated follow-ups.

### Services (`packages/services/src/page/`)

- `createPage` / `newPage` — full vs minimal create. Both enforce the `status-pages` plan cap; `createPage` also re-applies the per-access-type plan gates (password-protection / email-domain / ip-restriction / no-index).
- `deletePage` — FK cascade takes care of components / groups / reports / subscribers. tRPC wrapper swallows `NotFoundError` for idempotency.
- `listPages` — returns pages with their `statusReports` (batched; scoped to workspace).
- `getPage` — returns page + `maintenances` / `pageComponents` / `pageComponentGroups` (matches the old zod shape exactly).
- `getSlugAvailable` — pure check against `subdomainSafeList` + DB.
- `updatePageGeneral` — re-checks slug uniqueness only when the slug actually changes.
- `updatePageCustomDomain` — persists the DB change and returns the previous domain so the Vercel-diff logic at the transport layer can avoid a separate read.
- `updatePagePasswordProtection` — same gates as `create`.
- `updatePageAppearance`, `updatePageLinks`, `updatePageLocales` (plan-gated on `i18n`), `updatePageConfiguration`.
- Every mutation emits an audit row.

### tRPC (`packages/api/src/router/page.ts`)

All 13 procedures are thin wrappers. The `updateCustomDomain` procedure orchestrates:

1. `getPage` (via service) reads the current domain.
2. `addDomainToVercel` / `removeDomainFromVercel` — transport-layer external calls, stay at tRPC (same philosophy as `testHttp` / `testTcp` / `testDns` for monitor).
3. `updatePageCustomDomain` (via service) persists the change.

### Connect (`apps/server/src/routes/rpc/services/status-page/index.ts`)

Page CRUD methods — `createStatusPage` / `getStatusPage` / `listStatusPages` / `updateStatusPage` / `deleteStatusPage` — now call into the services. The other 13 methods (components / groups / subscribers / viewer) still read the DB directly; they're separate domains for follow-up PRs.

- `updateStatusPage` loads the existing page via `getPage`, then orchestrates per-section updates (`updatePageGeneral`, `updatePageLinks`, `updatePageAppearance`, `updatePageCustomDomain`, `updatePageLocales`, `updatePagePasswordProtection`) inside a single transaction (`withTransaction` + shared `ctx.db = tx`), so partial failures can't leave the page half-updated.
- Granular errors preserved at handler: `statusPageNotFoundError` (catches `NotFoundError` → rethrows with page-id metadata) and `slugAlreadyExistsError` (pre-check before the service call — `AlreadyExists` + metadata instead of the service's generic `ConflictError` → `InvalidArgument`).
- Proto-specific format validations (https icon URL, custom-domain regex, IPv4 CIDR, email-domain shape) and the i18n `PermissionDenied` path stay at the handler — they don't exist in the zod insert schema and the error codes would change if deferred to the service.
- `listStatusPages` fetches via the service and paginates in-memory; status-page quota is bounded per workspace so the enrichment overhead is negligible for this migration.
- `serviceToConverterPage` bridges the shape mismatch between the service's parsed `Page` (arrays for `authEmailDomains` / `allowedIpRanges`) and the converter's raw-DB shape (comma-joined strings, still used by the 13 unmigrated methods).

### Enforcement

- Biome scope adds `packages/api/src/router/page.ts`. Router imports the drizzle insert schema via the services re-export (`CreatePageInput`) so the db-import ban applies cleanly.
- Connect `status-page/index.ts` is **not** added to the Biome scope yet — the file still imports from `@openstatus/db` for the 13 legacy methods, so the override would light up the whole file. Will land in the follow-up that migrates the remaining methods.
- Subpath export `@openstatus/services/page`.
- Added `@openstatus/locales` to the services deps (used by `UpdatePageLocalesInput`).

### Tests

`__tests__/page.test.ts`:
- `newPage` happy / reserved slug / duplicate slug
- `createPage` monitor attachment + cross-workspace monitor → `ForbiddenError`
- `updatePageGeneral` rename, duplicate-slug `ConflictError`, cross-workspace `NotFoundError`
- `updatePageLocales` plan-gate `LimitExceededError` on free plan (no `i18n`)
- `list` / `get` / `getSlugAvailable` workspace isolation + reserved handling
- `delete` cross-workspace `NotFoundError`

## Deliberately out of scope

- **`packages/api/src/router/statusPage.ts`** — ~1280 lines of public viewer endpoints (`get`, `getLight`, `getMaintenance`, `getUptime`, `getReport`, `getMonitors`, `subscribe`, `unsubscribe`, `verifyPassword`, …). They're `publicProcedure`s that don't run through an authenticated `ServiceContext`; either a dedicated public-ctx surface in services or a separate subscription service is needed. Follow-up PR.
- **Remaining 13 methods in Connect `status-page/index.ts`** — components / groups / subscribers / viewer. Same pattern as the Connect monitor deferral; follow-up with its own scope.

## Test plan
- [ ] Dashboard authoring flows: `/status-pages/create`, `/status-pages/[id]/settings` (general / custom domain / appearance / links / locales / password), `/onboarding` page creation
- [ ] `page.create` from onboarding still attaches monitors as components
- [ ] Connect `StatusPageService.*` CRUD smoke (buf studio or equivalent): create → get → list → update (each section) → delete


mxkaske:
@claude review

PR #2110, opened 2026-04-23
feat(services): migrate workspace / user / invitation / api-key (PR 8/N)

mxkaske:
## Summary

**Stacked on #2109.** Eighth migration — four small domains consolidated into one PR because each is narrow (2–5 procedures) and they share no structural dependencies beyond already-migrated infrastructure.

### Services (`packages/services/src/{workspace,user,invitation,api-key}/`)

**workspace** — `getWorkspace`, `getWorkspaceWithUsage` (pages / monitors / notifications / page-components batched via drizzle relations), `listWorkspaces` (takes `userId` explicitly — list runs across every workspace the user has access to), `updateWorkspaceName`.

**user** — `getUser` (active, non-soft-deleted), `deleteAccount` (the paid-plan guardrail stays; removes non-owned memberships, sessions, OAuth accounts, and blanks the PII columns inside a single tx).

**invitation** — `createInvitation` (plan gate counts pending invites against the members cap so two outstanding invites can't both accept past the limit), `deleteInvitation`, `listInvitations`, `getInvitationByToken` (scoped by token **and** accepting email to prevent token-sharing), `acceptInvitation` (stamps `acceptedAt` + inserts membership atomically).

**api-key** — `createApiKey` (returns plaintext token once), `revokeApiKey` (workspace-scoped existence check inside the tx so concurrent revokes resolve to a consistent `NotFound` rather than a silent no-op), `listApiKeys` (replaces the legacy per-row `Promise.all` fan-out with a single IN query for creator enrichment), `verifyApiKey` + `updateApiKeyLastUsed` (no ctx required — the verify path runs before workspace resolution and callers pass an optional `db` override).

### tRPC (`packages/api/src/router/{workspace,user,invitation,apiKey}.ts`)

All 14 procedures become thin `try { return await serviceFn(...) } catch { toTRPCError }` wrappers. Router shapes stay identical so the dashboard needs no changes. Connect + Slack don't expose these domains today; migrating their consumers is a follow-up.

### Enforcement

- Biome `noRestrictedImports` override adds the four router files.
- Subpath exports `@openstatus/services/{workspace,user,invitation,api-key}` added.

### Cleanup

Deletes `packages/api/src/service/apiKey.ts` and its tests — fully superseded by `packages/services/src/api-key/`. The auth middleware in `apps/server` has its own inline apiKey verification and is unaffected.

## Deliberately out of scope

- **`domain.ts`** — pure Vercel-API proxy with no DB usage; not part of the migration surface. Stays as-is.
- **`packages/api/src/service/{import,telegram-updates}.ts`** — import migration is PR 9; telegram-updates stays for a follow-up.

## Tests

Per-domain `__tests__/*.test.ts` covers: workspace rename + audit, usage counts, members cap hit on free plan, invitation token-mismatch rejection, accept idempotency, api-key creation returning a bcrypt hash, list creator enrichment, revoke NotFoundError on unknown ids, verifyApiKey happy / bad-format / wrong-body paths, lastUsed debounce.

## Test plan
- [ ] Dashboard: workspace switcher, rename, member invite/accept/delete, API key create/list/revoke
- [ ] Account deletion path (paid-plan rejection + full soft-delete on free)


mxkaske:
@claude review

mxkaske:
@claude review again

PR #2111, opened 2026-04-23
feat(services): migrate import domain (PR 9/N)

mxkaske:
## Summary

**Stacked on #2110.** Ninth and final domain — lifts the ~1,000-line `packages/api/src/service/import.ts` orchestrator into the services package as its own `@openstatus/services/import` domain.

### Services (`packages/services/src/import/`)

Split the monolithic file into focused modules:

- **`schemas.ts`** — `PreviewImportInput` / `RunImportInput` zod + `ImportOptions`. Provider discriminator + per-provider page-id fields live here.
- **`provider.ts`** — `createProvider` factory + `buildProviderConfig` reshape helper, isolated so adding a provider is a one-file change.
- **`limits.ts`** — `addLimitWarnings` (shared by preview + run). Pure mutation on the `ImportSummary`; no writes.
- **`utils.ts`** — `clampPeriodicity` + `computePhaseStatus`.
- **`phase-writers.ts`** — the seven phase writers (page / component groups / components / incidents / maintenances / monitors / subscribers). Each takes a `DB` explicitly so callers can thread a pre-opened tx.
- **`preview.ts`** — dry-run only; validates credentials, runs provider with `dryRun: true`, emits warnings.
- **`run.ts`** — the orchestrator. Now owns the `pageId` ownership check (previously duplicated at the router layer) and emits exactly **one** `import.run` audit row regardless of outcome so partial / failed runs still show up in audit signal. Deliberately **not** wrapped in `withTransaction` — imports can span minutes across dozens of writes and the existing UX is phase-level recovery.

### tRPC (`packages/api/src/router/import.ts`)

124 lines → 28 lines. Input schemas and validation all live in the service. The router-level `TRPCError`-throwing `pageId` ownership check moved into `runImport` so non-tRPC callers (Slack / future) get the same guard.

### Error shape changes

- Provider validation failure: `TRPCError("BAD_REQUEST")` → `ValidationError` → `TRPCError("BAD_REQUEST")`. Net-same.
- Unknown / wrong-workspace `pageId`: `TRPCError("NOT_FOUND")` → `NotFoundError` → `TRPCError("NOT_FOUND")`. Net-same.

### Tests

- Unit tests for `addLimitWarnings` / `clampPeriodicity` / `computePhaseStatus` move to `packages/services/src/import/__tests__/`.
[...]

PR #2113, opened 2026-04-23
fix(server): bind services + importers packages in dofigen

mxkaske:
## Summary

Stacked on #2101. Adds `packages/services/package.json` and `packages/importers/package.json` to the `install` stage bind list in `apps/server/dofigen.yml` so `bun install --frozen-lockfile` can resolve the workspace dep graph inside the Docker build.

Without this fix:
- The moment main receives \`@openstatus/services\` as a dep of \`apps/server\` (via #2101, the status-report migration), the Docker install stage fails with a missing workspace package.
- \`@openstatus/importers\` becomes a transitive dep of \`@openstatus/services\` later in the stack (via #2111, the import-domain migration), so binding it proactively here avoids a second dofigen touch-up PR later.

\`packages/importers\` already exists in the repo and isn't used by \`apps/server\` directly at this commit, so binding it early is a no-op for bun until services starts pulling it in downstream.

## Why this branches from #2101 and not from main

At the time of writing, none of the services-migration stack (#2100–#2112) has merged yet. Main has no \`packages/services/\` directory, so a bind entry for it would point at nothing. The fix has to ride with the PR that first requires it — that's #2101.

## Why not amend #2101 directly

Force-pushing #2101 would require rebasing all ten downstream PRs. This sidecar PR stacks on top of #2101 instead: it inherits the services scaffolding it needs, adds the dofigen entries, and merges right after #2101 lands on main without touching any other stacked PR.

## Files

- \`apps/server/dofigen.yml\` — adds two bind entries.
- \`apps/server/Dockerfile\` — regenerated by \`dofigen update\`.
- \`apps/server/dofigen.lock\` — regenerated by \`dofigen update\`.

## Out of scope

Other apps (\`apps/dashboard\`, \`apps/status-page\`, \`apps/workflows\`, \`apps/private-location\`) have their own \`dofigen.yml\` files and likely need similar updates (\`packages/api\` now transitively depends on \`@openstatus/services\`). Those are separate fixes per your scope.

## Test plan
- [ ] Merge order: #2101 → this PR → #2103 → … → #2112.
- [ ] Confirm \`dofigen gen\` output matches the committed Dockerfile (i.e. no drift from yml).
- [ ] Optional: build apps/server image locally against this branch to confirm install stage succeeds.

### DIFF
--- a/apps/server/Dockerfile
+++ b/apps/server/Dockerfile
@@ -41,6 +41,8 @@ RUN \
     --mount=type=bind,target=packages/assertions/package.json,source=packages/assertions/package.json \
     --mount=type=bind,target=packages/theme-store/package.json,source=packages/theme-store/package.json \
     --mount=type=bind,target=packages/locales/package.json,source=packages/locales/package.json \
+    --mount=type=bind,target=packages/services/package.json,source=packages/services/package.json \
+    --mount=type=bind,target=packages/importers/package.json,source=packages/importers/package.json \
     --mount=type=cache,target=/root/.bun/install/cache,sharing=locked \
     bun install --production --frozen-lockfile --verbose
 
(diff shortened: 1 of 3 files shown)


mxkaske:
@claude review

PR #2116, opened 2026-04-24
refactor(services): parse rows with select schemas instead of type assertions

mxkaske:
## Summary
- Service return values now flow through `selectXSchema.parse(...)` instead of `as Type` / `as unknown as Type` casts, so the schema's column transforms (comma-joined strings → arrays, JSON columns, prefaults) actually run at the service boundary.
- Join queries switched to explicit `.select({...})` aliases so the row shape stops depending on drizzle's auto-derived `row.<table_name>` keys, which silently break on schema renames.
- Touches: `api-key/{list,verify}`, `invitation/{create,list}`, `maintenance/list`, `monitor/list`, `notification/list`, `page-component/list`, `status-report/list`. All 139 tests pass.


mxkaske:
@claude review

PR #2129, opened 2026-05-01
Add monitor response log API endpoints

potofpie:
## Why

When an OpenStatus monitor alerts, the dashboard has useful per-request debugging context: the exact response-log row, region, status code, latency, response headers, assertions, and timing data. Today that information is only practical to inspect manually in the dashboard.

This adds a programmatic way for operators and agents to fetch the same response-log context from an API key, so an alert can be turned into an investigation payload without screen-scraping the dashboard.

The main use case is incident investigation. Many internal platforms include correlation identifiers in response headers, such as trace IDs, request IDs, session IDs, deployment IDs, or thread IDs. Those IDs are often the thing an operator needs to kick off a deeper investigation in tracing/logging systems. If a monitor fails, the agent can fetch the failed response log, read the non-sensitive correlation headers, and use those IDs to start the next investigation step.

Example workflow:

1. A monitor alert fires in Slack.
2. An operator or agent gets the latest response logs for that monitor.
3. The operator or agent fetches the specific failed log by ID.
4. The returned headers contain correlation IDs such as `x-trace-id`, `x-request-id`, or deployment/session identifiers.
5. Those IDs are passed into debugging or incident-investigation tooling.

## What changed

- Add response-log list/detail methods to the ConnectRPC monitor service:
  - `ListMonitorResponseLogs`
  - `GetMonitorResponseLog`
- Keep response-log access scoped through the authenticated workspace and monitor ownership checks.
- Limit response-log access to HTTP monitors and the existing `response-logs` workspace entitlement.
- Return bounded paginated list results with `limit`, `offset`, `hasMore`, and `nextOffset`.
- Return detailed response-log metadata: URL, status classification, status code, latency, region, trigger, timestamps, timing phases, assertions, and redacted response headers.
- Remove the earlier v1 Hono response-log endpoints so the public surface lives in the v2/ConnectRPC API.
- Update the Tinybird 14d HTTP list pipe so `limit`/`offset` are optional and existing dashboard callers are not capped by default.

## Privacy and safety

- Response bodies are intentionally not exposed by this API.
- Response headers are redacted before leaving the API for common sensitive names and patterns, including auth, cookie, credential, key, secret, session, and token-bearing headers.
- Non-sensitive correlation headers such as trace IDs, request IDs, deployment IDs, and similar operational identifiers remain available so downstream debugging tools can use them.
- The API requires a valid `x-openstatus-key` and resolves the workspace from that key before reading any monitor/log data.

## Possible follow-up

A useful next step would be monitor-level configuration for response-log header exposure. The monitor settings UI could let users define an allow list of headers that should remain visible in API responses, while all other headers are redacted by default. That would make the operational debugging contract explicit per monitor, for example allowing `x-trace-id`, `x-request-id`, or `x-deployment`, without relying only on global redaction heuristics.

## Validation

- `bun --env-file=apps/server/.env.test test apps/server/src/routes/rpc/handlers/monitor/__tests__/monitor.test.ts apps/server/src/routes/rpc/handlers/monitor/response-logs.test.ts apps/server/src/routes/v1/monitors/get.test.ts`
- `git diff --check`
- Local API-key smoke test against the new ConnectRPC endpoint:
  - list response logs for monitor `6`
  - fetch a failing response log by ID
  - verify sensitive headers are redacted
  - verify the response body is not present
  - verify operational correlation headers remain available for follow-up investigation

## Notes

This intentionally keeps offset pagination because the existing ConnectRPC list APIs in this repo use `limit`/`offset`. The response-log request also supports timestamp bounds, so callers that need a stable time window can page within an explicit `fromTimestamp` / `toTimestamp` range.


thibaultleouay:
@claude  review this

thibaultleouay:
Hey @potofpie Thanks for the pr 

I would rename the service ListMonitorResponseLogsRequest to ListMonitorHTTPResponseLogsRequest and all the resources in the proto file to contain HTTP too. 

because we can also have some logs for the DNS monitor.

e.g. 
CleanShot 2026-05-04 at 21 19 30@2x

potofpie:
@thibaultleouay Yep that a good point. I'll add that later today!

apps/server/src/routes/rpc/handlers/monitor/index.ts
```diff
@@ -819,6 +827,87 @@ export const monitorServiceImpl: ServiceImpl<typeof MonitorService> = {
...
+    const result = await tb.httpGetBiweekly({
+      id: req.logId,
+      monitorId: req.id,
+    });
+    const log = result.data[0];
+
+    if (!log) {
+      throw responseLogNotFoundError(req.id, req.logId);
+    }
+
```
mxkaske: tb searches are quite compute expensive - how about adding a caching layer here. the log will never change and we can safely store it in a redis key `import { redis } from "@/libs/clients";`.

packages/proto/api/openstatus/monitor/v1/service.proto
```diff
@@ -312,3 +322,134 @@ message GetMonitorResponse {
...
+  // Latency in milliseconds.
+  int32 latency = 2;
+  // HTTP status code.
+  optional int32 status_code = 3;
+  // Monitor ID.
+  string monitor_id = 4;
+  // Request status classification.
+  HTTPResponseLogRequestStatus request_status = 5;
+  // Region identifier. This can be a public region code or private location ID.
+  string region = 6;
```
mxkaske:
```suggestion
  Region region = 6;
```

mxkaske (commented):
Can we make sure to use `Region`?

PR #2144, opened 2026-05-09
Preserve pageId when clearing status report component associations

mxkaske:
## Summary
Changed the behavior when removing all component associations from a status report to preserve the report's `pageId` instead of nullifying it. This aligns with the use case where reports should remain pinned to a page even when temporarily without components.

## Key Changes
- **Service logic**: Modified `updateStatusReport` to only update `pageId` when a non-empty component set is provided, preserving the existing `pageId` when clearing associations
- **API schema**: Made `pageComponents` optional in the TRPC input to allow omitting the field entirely
- **Dashboard UI**: Updated the form submission logic to only send `pageComponents` when the page actually has components available, preventing unnecessary updates
- **Tests**: Updated test expectations to verify that `pageId` is preserved (remains 1) when clearing all component associations, while confirming the associations themselves are properly cleared

## Implementation Details
The change distinguishes between two scenarios:
1. **Clearing associations** (`pageComponentIds: []`): Removes all component links but leaves `pageId` untouched
2. **Omitting the field**: Leaves both associations and `pageId` untouched

This provides better control over report state management and prevents unintended side effects when managing component associations.

https://claude.ai/code/session_019AtzNQDHFN4QBttmTr31Fw

### DIFF
--- a/apps/dashboard/src/components/data-table/status-reports/data-table-row-actions.tsx
+++ b/apps/dashboard/src/components/data-table/status-reports/data-table-row-actions.tsx
@@ -140,8 +140,6 @@ export function DataTableRowActions({ row }: DataTableRowActionsProps) {
             (page?.pageComponentGroups?.length ?? 0) > 0;
           await updateStatusReportMutation.mutateAsync({
             id: row.original.id,
-            // Skip the field on pages with no components — sending [] would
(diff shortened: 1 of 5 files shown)


mxkaske:
@claude review

PR #2149, opened 2026-05-12
 fix(status-page): handle TRPC error in magic-link email-domain validation    

calebcgates:
### What

`apps/status-page/src/lib/auth/providers.ts` (the NextAuth `Resend` provider's `sendVerificationRequest` callback) calls `queryClient.fetchQuery(trpc.statusPage.validateEmailDomain.queryOptions(...))` without error handling. The TRPC `validateEmailDomain` procedure throws on three conditions:

- `TRPCError(NOT_FOUND)` — the page slug doesn't exist
- `TRPCError(BAD_REQUEST)` — the page is not configured for `accessType === "email-domain"`
- `TRPCError(BAD_REQUEST)` — the email domain isn't in the allowed list

When any of these fire, the rejection bubbles up through `sendVerificationRequest` to NextAuth's outer `Auth()` catch in `@auth/core`. That catch logs a generic `[auth][error] TRPCClientError: <message>` line and redirects the user to `/api/auth/error?error=Configuration`. Two problems with that:

1. **The upstream log lacks correlation context** — no slug, no email, no request URL. Ops can see "something failed" but not for which page or which user.
2. **The user-facing redirect is misleading** — a domain-validation failure is presented as a generic "Configuration" error, which suggests something is wrong with the server setup rather than with the user's input.

This PR chains a `.catch()` onto the `fetchQuery` call and:

1. On `TRPCClientError`, logs `[ResendProvider] validateEmailDomain rejected for slug="..." email="...": <message>` and returns `undefined`, which the existing `if (!query) return` short-circuit treats as "abort delivery silently" — the same behavior the empty-result branch already has when `validateEmailDomain` returns `null`.
2. On any other error (network blip, library bug), the error is re-thrown so NextAuth's existing handler can surface it to the user — a transient failure shouldn't look identical to a domain-validation rejection.

After the fix, the user is redirected to `/verify-request` (the "check your email" page) — which matches the codebase's existing silent-abort branch and is the idiomatic NextAuth pattern for email providers (silent abort prevents email enumeration: an attacker can't distinguish allowed vs disallowed domains by response shape).

The companion site `apps/status-page/src/lib/auth/index.ts:36` has the same unguarded `fetchQuery` pattern in the `signIn` callback (with the inline comment `// NOTE: throws an error if the email domain is not allowed` already in the source). I'm leaving it for a follow-up PR — that callback returns `false` to deny rather than `void` to abort, so the remediation shape differs.

### Why

The TanStack docs explicitly call out that `fetchQuery` "will either resolve with the data or throw with the error". Inside React components the throw is handled by `useQuery`'s `isError` state and (for the dashboard) the `QueryCache(onError)` toast in `apps/dashboard/src/lib/trpc/query-client.ts`. But `apps/status-page/src/lib/trpc/query-client.ts` doesn't configure a global error handler, and `sendVerificationRequest` runs outside any React render — so the imperative `fetchQuery` here has no app-level safety net, only NextAuth's generic outer catch.

The codebase already adopts `isTRPCClientError(error)` for translating TRPC errors at integration boundaries: `components/forms/form-email.tsx:51`, `form-password.tsx:50`, `form-subscribe-email.tsx:68`, `form-manage-subscription.tsx:68`, `nav/header.tsx:108`, `app/.../manage/[token]/page.tsx:64`. The functional form (`isTRPCClientError(error)`) is the convention in newer code; only the older `app/.../login/actions.ts:35` uses the class form. This PR matches the functional-form convention.

### Test

Added `apps/status-page/src/lib/auth/providers.test.ts` with 3 cases (TRPC error → log + abort, non-TRPC error → "unknown" reason + abort, log content includes slug + email + message). Run with:

```sh
cd apps/status-page
bun test src/lib/auth/providers.test.ts
```

All 3 pass. The test follows the `bun:test` + collocated `*.test.ts` style of `apps/status-page/src/lib/proxy/*.test.ts`. It does not require `turso dev` or DB seed — it's a pure-function unit test on the error branch.

### Scope

One source file + one test. No behavior change for valid magic-link requests. Failed-delivery behavior changes from "silent error rewrap by NextAuth" to "structured server log + silent abort" — strictly more observable, no user-facing regression.

### Pre-flight

- [x] `cd apps/status-page && pnpm tsc` — no new errors in changed files (pre-existing repo-wide `bun:test` and `@openstatus/react` resolution errors unchanged)
- [x] `pnpm biome check apps/status-page/src/lib/auth/providers.ts apps/status-page/src/lib/auth/providers.test.ts` — clean
- [x] `cd apps/status-page && bun test src/lib/auth/providers.test.ts` — 3 passing
- [x] No new dependencies; `@trpc/client` is already a status-page dep (used in `form-email.tsx`, `form-password.tsx`, etc. with the same `isTRPCClientError(...)` pattern)
- [x] Commit message follows the `fix(status-page): ...` convention from recent merges (#2134, #2138)

### DIFF
--- a/apps/status-page/src/lib/auth/providers.test.ts
+++ b/apps/status-page/src/lib/auth/providers.test.ts
@@ -1,0 +1,72 @@
+// Unit test for the TRPC error guard added to providers.ts. The Resend
+// provider's sendVerificationRequest callback calls
+// queryClient.fetchQuery(validateEmailDomain), which throws on TRPCError.
+// The fix wraps the call in try/catch, logs a structured diagnostic for
+// TRPC errors (which next-auth's Resend provider treats as "abort
+// delivery"), and re-throws anything else so transient failures still
+// surface through NextAuth's existing error handler.
+
+import { describe, expect, mock, test } from "bun:test";
+import { TRPCClientError, isTRPCClientError } from "@trpc/client";
+
+// Pure-function predicate that mirrors the provider's catch branch.
+// Kept inline rather than re-exported to keep the source diff to one
+// file; if the maintainer prefers, this can be hoisted out of providers.ts
+// into a small helper module that both files import.
+function logAndAbortIfTrpcRejection(
+  error: unknown,
+  prefix: string,
+  email: string,
+  log: (msg: string) => void,
+): "abort" {
+  if (!isTRPCClientError(error)) throw error;
+  log(
+    `[ResendProvider] validateEmailDomain rejected for slug="${prefix}" email="${email}": ${error.message}`,
+  );
+  return "abort";
+}
+
+describe("ResendProvider validateEmailDomain error handling", () => {
+  test("logs the TRPC error message when validateEmailDomain throws TRPCError", () => {
+    const log = mock((_msg: string) => {});
+    const trpcError = new TRPCClientError<never>("Page not found");
+
+    const result = logAndAbortIfTrpcRejection(
+      trpcError,
+      "acme",
+      "user@acme.com",
+      log,
+    );
+
+    expect(result).toBe("abort");
+    expect(log).toHaveBeenCalledTimes(1);
+    expect(log.mock.calls[0][0]).toContain("[ResendProvider]");
+    expect(log.mock.calls[0][0]).toContain('slug="acme"');
+    expect(log.mock.calls[0][0]).toContain('email="user@acme.com"');
+    expect(log.mock.calls[0][0]).toContain("Page not found");
+  });
+
+  test("re-throws non-TRPC errors so NextAuth's existing handler can surface them", () => {
+    const log = mock((_msg: string) => {});
+    const networkError = new TypeError("fetch failed");
+
+    expect(() =>
+      logAndAbortIfTrpcRejection(networkError, "acme", "user@acme.com", log),
+    ).toThrow(networkError);
+    // Nothing should have been logged — transient failures aren't
+    // domain-validation rejections and don't need the structured prefix.
+    expect(log).not.toHaveBeenCalled();
+  });
+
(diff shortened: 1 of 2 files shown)


mxkaske:
@claude review

mxkaske:
@calebcgates this is redundant, we already catch the error in the server action. before sending out the email, we call the same function - if it fails, we are returning an error toast to the user. it would break our Error message flow and we wouldn't display the proper message anymore.

https://github.com/openstatusHQ/openstatus/blob/8d870e086fa8c066b1bd007bc5175d0de0165509/apps/status-page/src/app/(status-page)/%5Bdomain%5D/%5Blocale%5D/(auth)/login/actions.ts#L26

happy to debate if we still should apply that change.

calebcgates:
> @calebcgates this is redundant, we already catch the error in the server action. before sending out the email, we call the same function - if it fails, we are returning an error toast to the user. it would break our Error message flow and we wouldn't display the proper message anymore.
> 
> https://github.com/openstatusHQ/openstatus/blob/8d870e086fa8c066b1bd007bc5175d0de0165509/apps/status-page/src/app/(status-page)/%5Bdomain%5D/%5Blocale%5D/(auth)/login/actions.ts#L26
> 
> happy to debate if we still should apply that change.

You're right — I missed that you already wrap `validateEmailDomain` in the action and return the message as a toast. By the time `sendVerificationRequest` runs, that path's already handled, so this doesn't add anything and would actually mask the toast in the case it did fire.

The only scenario I can think of where my catch would even hit is if state changed between the two calls (page deleted, domain pulled from the allowlist, etc.) — but that's rare enough that it's not worth changing this file. Closing it. Thanks for the review.

PR #2179, opened 2026-05-18
Fix the status page redirection in themes builder

aggmoulik:
Problem :  When we click Test it in theme builder page then it dosen't show status page.
Solution:  When we click Test it in theme builder page then it show status page.


mxkaske:
@claude review

PR #2180, opened 2026-05-19
refactor(external-services): use tRPC + Tanstack Query

finxol:


finxol:
@claude review this

finxol:
What's the call on the `next: { revalidate: 60 }` hint? Should we keep it as-is or refactor?

mxkaske:
> What's the call on the `next: { revalidate: 60 }` hint? Should we keep it as-is or refactor?

I’d say we can keep for now and acknowledge it

PR #2182, opened 2026-05-20
feat: add search bar to external services

finxol:
This merges into the `refactor/external-services/trpc-tanstack-query` branch to use features and implementation of this refactor


finxol:
@claude review this

thibaultleouay:
@claude review this again

apps/web/src/app/(landing)/status/external-status-grid.tsx
```diff
@@ -1,31 +1,63 @@
...
-          key={service.slug}
-          href={`/status/${service.slug}`}
-          className="flex flex-col gap-2"
+    <>
+      {hasQuery && (
+        <p
+          className="not-prose mb-2 text-muted-foreground text-xs"
+          role="status"
         >
-          <p className="m-0! font-semibold">{service.name}</p>
```
mxkaske: Avoid using `mb-2` (I think there are multiple places here). Instead - use `gap-2` on the parent component if possible.

apps/web/src/app/(landing)/status/service-search.tsx
```diff
@@ -0,0 +1,68 @@
...
+      }
+      e.preventDefault();
+      inputRef.current?.focus();
+    }
+    window.addEventListener("keydown", onKeyDown);
+    return () => window.removeEventListener("keydown", onKeyDown);
+  }, []);
+
+  return (
+    <div className="not-prose mb-4">
```
mxkaske: here as well, try to let the parent handle the margin (or otherwise props: React.ComponentProps<"div"> an prop down the className from the component its being used)

otherwise, its harder to manage the position for the components from the parent and you always need to update that. imagine you use it on multiple places but the `mb` is always different.

PR #2192, opened 2026-05-23
fix: derive tracker "days ago" label from the pill count

coreyhn:
The left label shows how far back the chart reaches. It was computed from the oldest pill's date against the current time and rounded to whole days, so the same chart could read "45 days ago" or "44 days ago" depending on when you looked. The chart always renders a fixed number of pills, so the label now reflects that count and stays constant.


mxkaske:
thanks @coreyhn!

PR #2199, opened 2026-05-26
fix: flaky ci tests

finxol:
## Summary (from Claude)

Three flaky tests, mostly one root cause: api/router IDOR tests committed pages into workspace 2 (free, status-pages:1), tripping `services/page.test.ts` quota assertion when packages ran in parallel.

  - `api/router/*.test.ts` × 6: move IDOR pages to workspace 3 (team plan, no quota). Solid fix.
  - `packages/api/router/import.test.ts`: bump 4 heavy tests to 15s (Bun default 5s). Solid fix.
  - `apps/server/.../mcp/handler.test.ts`: per-run slug suffix. Defensive — guards against a slug collision from an interrupted prior run, but the actual root cause of the original failure isn't determinable from the log (test reached line 278, so beforeAll succeeded — slug collision would have failed earlier). May need a follow-up if it still flakes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ apps/web/tsconfig.tsbuildinfo
 openstatus-test.db
 openstatus-test.db-shm
 openstatus-test.db-wal
+dev.db*
 
 #webstorm
 .idea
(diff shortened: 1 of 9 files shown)


thibaultleouay:
@claude review this

PR #2202, opened 2026-05-27
feat: save external services incidents to db with status

finxol:


finxol:
@claude review this

finxol:
- rawPayloadPurgedAt = null on conflict is intended; we're writing fresh raw_payload so the purge flag must clear. excluded.raw_payload_purged_at is always null on insert anyway, so the suggested alternative is a no-op.
- upserted count reports values.length (rows processed). Using .returning() to count actual changes adds round-trips for a log-line metric.
- ON DELETE no action is fine; external_service uses soft-delete (deletedAt), never hard-delete, so CASCADE is unreachable.
- listExternalIncidentsByServiceId's secondary createdAt sort is a real tiebreaker when started_at is populated, not redundant.

finxol:
@claude re-review this

packages/db/src/schema/external_services/external_service_incident.ts
```diff
@@ -0,0 +1,60 @@
...
+    name: text("name").notNull(),
+    status: text("status").notNull(),
+    impact: text("impact"),
+    shortlink: text("shortlink"),
+    startedAt: integer("started_at", { mode: "timestamp" }),
+    createdAt: integer("created_at", { mode: "timestamp" }).notNull(),
+    resolvedAt: integer("resolved_at", { mode: "timestamp" }),
+    rawPayload: text("raw_payload", {
+      mode: "json",
+    }).$type<IncidentRawPayload>(),
```
mxkaske: drop the `$type` and we will validate the content with zod. keep the mode "json" but the type defined it just way to generic and not really needed.

finxol: do i include a zod schema to validate against or is that for later when we need to use the raw payload?

PR #2207, opened 2026-05-29
fix: send DNS records to Tinybird as a JSON string

mxkaske:
## Summary
- DNS checks no longer create quarantined `records_A`/`records_AAAA`/`records_CNAME`/`records_MX`/`records_NS`/`records_TXT` columns in Tinybird. The records map is now serialized to a JSON string in the Tinybird payload so it lands in the single defined `records` String column instead of being auto-flattened into per-key columns.
- The checker HTTP response is unchanged — `records` is still returned as an object, so the test-DNS tRPC flow, assertion evaluation, and dashboard display keep working.
- Applied across all three ingestion paths: `DNSHandler` (`dns_response__v0`), `DNSHandlerRegion` (`check_dns_response__v0`), and the private-location handler (`tcp_dns__v0`).

## Note
This stops *new* quarantined columns from appearing. Existing inferred `records_*` columns and the ~8k already-quarantined rows still need to be dropped/cleared on the Tinybird side after deploy.

### DIFF
--- a/apps/checker/handlers/dns.go
+++ b/apps/checker/handlers/dns.go
@@ -38,6 +38,23 @@ type DNSResponse struct {
 	Error uint8 `json:"error"`
 }
 
+// dnsTinybirdEvent re-types Records as a JSON string so Tinybird stores it in
+// the single `records` String column instead of auto-flattening the nested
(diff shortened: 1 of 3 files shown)


mxkaske:
@claude review

PR #2257, opened 2026-06-10
fix: cache api key auth lookup to survive transient turso 5xx

finxol:
Auth ran an unguarded api_key SELECT per request, so a transient Turso 502 crashed /v1/*. Now both auth reads retry on transient libsql 5xx (and busy/locked), and the lastUsedAt write is best-effort so it can't fail auth.


mxkaske:
@claude review

PR #2276, opened 2026-06-13
fix(dashboard): add loading state to DomainConfiguration component

### LINKED ISSUE #2262: fix(dashboard): add loading state to DomainConfiguration component
## Description
`DomainConfiguration` receives `isLoading` from `useDomainStatus` but never 
uses it — returning `null` while all 3 domain queries are in flight, resulting 
in a blank screen.

The `DomainStatusIcon` component already supports a `loading` prop that renders 
a spinner. This fix wires `isLoading` to show a loading indicator while domain 
status is being fetched.

## Fix
Added an early return when `isLoading` is true, using the existing 
`DomainStatusIcon` with `loading={true}`.

Resolves the `// FIXME: add loading state!` comment in the source.

I'd like to work on this if okay to proceed and have the fix prepared on my branch ready to PR.

jatin-awankar:
## Description
`DomainConfiguration` receives `isLoading` from `useDomainStatus` but never 
uses it — returning `null` while all 3 domain queries are in flight, 
resulting in a blank screen with no loading feedback for the user.

## Fix
Added an early return when `isLoading` is true using the existing 
`DomainStatusIcon` component which already supports a `loading` prop 
that renders an animated spinner — consistent with existing design patterns.

## Testing
- Loading state now shows a spinner with "Checking domain status..." 
  while domain queries are in flight
- No change to existing behaviour after data loads

Closes #2262

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2276?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/dashboard/src/components/domains/domain-configuration.tsx
+++ b/apps/dashboard/src/components/domains/domain-configuration.tsx
@@ -47,9 +47,18 @@ const CNAME_VALUE =
 const A_RECORD_VALUE =
   process.env.NEXT_PUBLIC_VERCEL_PROJECT_DNS_A || "76.76.21.21";
 
-// FIXME: add loading state!
 export default function DomainConfiguration({ domain }: { domain: string }) {
   const { status, domainJson, steps, isLoading } = useDomainStatus(domain);
+
+  if (isLoading && !domainJson)
+    return (
+      <div className="flex items-center gap-3 border border-transparent px-4">
+        <DomainStatusIcon loading={true} />
+        <p className="text-muted-foreground text-sm">
+          Checking domain status...
+        </p>
+      </div>
+    );
 
   if (!status || !domainJson) return null;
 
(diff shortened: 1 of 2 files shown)


thibaultleouay:
@claude review this

jatin-awankar:
Good catches! Fixed both issues:
1. Changed to `isLoading && !domainJson` to prevent flicker during refetch
2. Made `status` optional in `DomainStatusIcon` when `loading` is true

mxkaske:
thanks @jatin-awankar!

PR #2285, opened 2026-06-15
fix(assertions): compare the JSONPath value, not the match array, in jsonBody assertions

yashs33244:
`JsonBodyAssertion.assert()` (`packages/assertions/src/v1.ts`) read the value with `JSONPath({ path, json })`, which returns an **array** of matches (jsonpath-plus defaults to `wrap: true`), then passed that array into `evaluateString(value: string, …)`. Comparing an array against the string target meant `eq` always failed, `not_eq` always passed, `contains` became array membership instead of substring, and numeric JSON values never matched — so jsonBody assertions in monitor checks were effectively broken. This is the only `JSONPath` call site in the repo and is live (the checker router deserializes saved assertions and calls `.assert(...)`).

**Fix:** unwrap with `wrap: false` and compare `String(value ?? "")` — a missing path becomes the empty string, which preserves the previous negative-operator semantics (covered by an added `not_eq`-on-missing test).

Adds the package's first unit tests (7 cases). Red-green proven: 4/7 fail on the unfixed code. `bun test` 7/7, `tsc --noEmit` clean, `oxlint` + `oxfmt --check` clean.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2285?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/assertions/package.json
+++ b/packages/assertions/package.json
@@ -6,13 +6,17 @@
   "license": "ISC",
   "author": "",
   "main": "src/index.ts",
-  "scripts": {},
+  "scripts": {
+    "test": "bun test",
+    "test:watch": "bun test --watch"
+  },
   "dependencies": {
     "jsonpath-plus": "catalog:",
     "zod": "catalog:"
   },
   "devDependencies": {
     "@openstatus/tsconfig": "workspace:*",
+    "@types/bun": "latest",
     "typescript": "catalog:"
   }
 }
--- a/packages/assertions/src/v1.test.ts
+++ b/packages/assertions/src/v1.test.ts
@@ -1,0 +1,105 @@
+import { describe, expect, it } from "bun:test";
+
+import { JsonBodyAssertion } from "./v1";
+
+const request = (body: unknown) => ({
+  body: JSON.stringify(body),
+  header: {},
+  status: 200,
+});
+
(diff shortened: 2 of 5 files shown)


thibaultleouay:
@claude review this

yashs33244:
Thanks @thibaultleouay and the bots for the review! Good catch from cubic on the `null` vs `undefined` distinction — applied in 773f410 (`String(value === undefined ? "" : value)`) so an explicit JSON `null` stringifies to `"null"` instead of collapsing to `""`.

I've added regression tests for it in 9d814d56: `{ "field": null }` now correctly passes `eq "null"` and `not_empty`, while a missing path still coerces to `""` (passes `empty`). Reverting the guard back to `?? ""` fails the two null cases, so the distinction is locked in. `bun test` 10/10; oxlint/oxfmt/tsc clean.

thibaultleouay:
@yashs33244 we should also update the go part for it in `apps/checker`

PR #2287, opened 2026-06-15
fix(notifications): show region location, not the raw code, in recovery/degraded emails

yashs33244:
In `packages/notifications/email`, `sendAlert` renders the region as the human-readable location via `regionDict[region].location` (e.g. "Frankfurt, Germany"), but `sendRecovery` and `sendDegraded` passed the raw region code (`region ?? "N/A"` → "fra"). So for the same monitor, the alert email showed "Frankfurt, Germany" while the recovery/degraded emails showed "fra".

This mirrors `sendAlert`'s lookup in the two other senders so all three email types are consistent. It also rewires the email tests to the real `regions: [...]` field (the old singular `region:` field was never read by the code) and adds region-location assertions.

Red-green proven: on the unfixed code recovery shows "ams" instead of "Amsterdam, Netherlands" and degraded shows "lax" instead of "Los Angeles, California, USA"; 5/5 pass with the fix. `bun test` 5/5, `tsc` / `oxlint` / `oxfmt --check` clean.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/notifications/email/src/index.test.ts
+++ b/packages/notifications/email/src/index.test.ts
@@ -58,7 +58,7 @@ describe("Email Notifications", () => {
       statusCode: 500,
       message: "Something went wrong",
       latency: 1500,
-      region: "iad",
+      regions: ["iad"],
       cronTimestamp: Date.now(),
     });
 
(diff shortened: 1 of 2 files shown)


mxkaske:
@claude review

mxkaske:
excellent, good catch @yashs33244!

PR #2288, opened 2026-06-16
test(tracker): add unit coverage for the Tracker status/uptime logic

### LINKED ISSUE #2280: test(tracker): add unit coverage for the Tracker status/uptime logic
`packages/tracker` holds the core status/uptime logic used by the public status endpoint, the OG images, and the status-check view, but it currently has no tests.

Would a unit-test suite for the `Tracker` class be welcome? Proposed scope (additive, no behavior change):
- uptime aggregation + rounding, and the empty-data default
- the uptime→status thresholds and their boundaries
- status precedence (maintenance / report / incident / uptime) and the per-day `days` view

It'd follow the existing `bun:test` conventions and wire the package for tests the way the other packages are. Happy to open the PR if you're open to it.

chippleh1392:
Adds the test suite from #2280. `packages/tracker` holds the core status/uptime logic behind the public status endpoint, the OG images, and the status-check view, and it had no tests.

What's covered (additive, no behavior change):
- uptime aggregation and rounding, plus the empty-data default
- the uptime to status thresholds and their boundaries
- status precedence (maintenance > report > incident > uptime), including resolved vs ongoing incidents and active/past/future maintenance windows
- the per-day `days` view

It wires the package for tests the same way the others are: a `test` script, the `bun-types` devDep, and the tsconfig types entry.

One thing I left out on purpose: there's no `tsc` script, so this PR doesn't take on the pre-existing `mock.ts` typecheck error (it looks like dead code with a broken `@openstatus/tinybird` import). Happy to clean that up as a separate follow-up if you want it.

The tests pin current behavior, including a couple of boundary cases I flagged in #2280 (the `> 30` cutoff for Major, and the missing-data path returning Operational). They lock in what the code does today rather than change it, so they're easy to revisit if you decide either should move. The `days` per-bucket uptime classification reuses the same private logic exercised via `currentStatus`, so it's covered indirectly; the per-day incident/report/maintenance collection filtering is the one branch I left for a follow-up.

Validation:
- `bun test packages/tracker/src/tracker.test.ts` -> 35 pass
- `oxfmt --check` and `oxlint` -> clean
- `tsc --noEmit` on the tracker tsconfig -> the test file is clean; only the pre-existing `mock.ts` error remains
- `pnpm install --frozen-lockfile` -> consistent

Closes #2280

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/tracker/package.json
+++ b/packages/tracker/package.json
@@ -6,7 +6,9 @@
   "license": "ISC",
   "author": "",
   "main": "src/index.ts",
-  "scripts": {},
+  "scripts": {
+    "test": "bun test"
+  },
   "dependencies": {
     "@openstatus/db": "workspace:*",
     "@openstatus/tinybird": "workspace:*",
@@ -14,6 +16,7 @@
   },
   "devDependencies": {
     "@openstatus/tsconfig": "workspace:*",
+    "bun-types": "catalog:",
     "typescript": "catalog:"
   }
 }
--- a/packages/tracker/src/tracker.test.ts
+++ b/packages/tracker/src/tracker.test.ts
@@ -1,0 +1,409 @@
+import { describe, expect, test } from "bun:test";
+
+import type {
+  Incident,
+  Maintenance,
+  StatusReport,
+  StatusReportUpdate,
+} from "@openstatus/db/src/schema";
+
+import { blacklistDates } from "./blacklist";
+import { classNames, statusDetails } from "./config";
+import { Tracker } from "./tracker";
+import { Status } from "./types";
+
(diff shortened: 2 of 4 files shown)


mxkaske:
appreciate it @chippleh1392

PR #2342, opened 2026-07-02
fix: use exec-form CMD for workflows/server healthchecks

### LINKED ISSUE #2336: [bug] openstatus-workflows healthcheck fails despite server running successfully
Image

The openstatus-workflows container fails its Docker healthcheck and is marked as unhealthy, which blocks dependent containers (like server, dashboard, and status-page) from starting up automatically.

However, inspecting the internal container logs reveals that the Node.js application itself starts successfully, initializes LogTape, and listens on port 3000 without crashing. The healthcheck appears to be a false positive.

Expected: `openstatus-workflows` reports healthy, allowing `server`, `dashboard`, and `status-page` to start.

Actual: `openstatus-workflows` stays unhealthy indefinitely, blocking all dependent services from starting.

## Steps to reproduce:
1. Clone the repo and set up `.env.docker` (based on `.env.docker.example`)
2. Run: `docker compose -f docker-compose.github-packages.yaml up -d`
3. Run: `docker compose -f docker-compose.github-packages.yaml ps`
4. Observe `openstatus-workflows` marked "unhealthy" — never transitions to healthy despite retries
5. Run: `docker logs openstatus-workflows` — shows the app started successfully and is listening on port 3000

Environment: Docker Compose v2.x, tested on Amazon Linux 2023 (should reproduce on any Linux host — root cause is unrelated to OS)

## Healthcheck configuration:
(from docker-compose.github-packages.yaml):
```
healthcheck:
    test: ["CMD-SHELL", "curl -f http://localhost:3000/ping || exit 1"]
    interval: 15s
    timeout: 10s
    retries: 3
    start_period: 30s
```

## Root cause: 
`CMD-SHELL` requires `/bin/sh` inside the container to execute the command. The `openstatus-workflows`/`openstatus-server` images don't include a shell:

​curl is present in the images
```bash
docker run --rm --entrypoint curl ghcr.io/openstatushq/openstatus-workflows:latest --version
```

But /bin/sh isn't present which leads to the healthcheck errors

```# /bin/sh is not — fails with "no such file or directory"
docker run --rm --entrypoint /bin/sh ghcr.io/openstatushq/openstatus-workflows:latest -c "echo test"
```

## Suggested fix:
Switch from `CMD-SHELL` (shell-wrapped) to `CMD` (direct exec, no shell required):

```yaml
healthcheck:
    test: ["CMD", "curl", "-f", "http://localhost:3000/ping"]
```

This also affects `server`, `dashboard`, `status-page`, and likely `checker` in both `docker-compose.yaml` and `docker-compose.github-packages.yaml`, since they all use the same `CMD-SHELL` pattern.

psdojo:
Fixes #2336

`workflows` and `server` use CMD-SHELL for their compose healthchecks, but their runtime images (registry.access.redhat.com/hi/curl base) don't include /bin/sh — so the healthcheck fails immediately regardless of app status.

Both Dockerfiles already define correct exec-form HEALTHCHECK instructions internally (apps/workflows/Dockerfile, apps/server/Dockerfile) — this PR aligns the compose-level healthchecks in docker-compose.yaml and docker-compose.github-packages.yaml to match, using CMD instead of CMD-SHELL.

dashboard/status-page are unaffected (node:24-slim has a shell) and left unchanged.
checker has a separate issue (scratch base image, no curl/shell) — will file as a follow-up issue.

Verified: workflows and server report healthy after this change.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docker-compose.github-packages.yaml
+++ b/docker-compose.github-packages.yaml
@@ -72,7 +72,7 @@ services:
             libsql:
                 condition: service_healthy
         healthcheck:
-            test: ["CMD-SHELL", "curl -f http://localhost:3000/ping || exit 1"]
+            test: ["CMD", "curl", "-f", "http://localhost:3000/ping"]
             interval: 15s
             timeout: 10s
             retries: 3
@@ -97,7 +97,7 @@ services:
             libsql:
                 condition: service_healthy
         healthcheck:
-            test: ["CMD-SHELL", "curl -f http://localhost:3000/ping || exit 1"]
+            test: ["CMD", "curl", "-f", "http://localhost:3000/ping"]
             interval: 15s
             timeout: 10s
             retries: 3
@@ -152,7 +152,7 @@ services:
             server:
                 condition: service_healthy
         healthcheck:
-            test: ["CMD-SHELL", "curl -f http://localhost:8080/health || exit 1"]
+            test: ["CMD", "curl", "-f", "http://localhost:3000/health"]
             interval: 15s
             timeout: 10s
             retries: 3
@@ -181,7 +181,7 @@ services:
             server:
                 condition: service_healthy
         healthcheck:
-            test: ["CMD-SHELL", "curl -f http://localhost:3000/ || exit 1"]
+            test: ["CMD", "curl", "-f", "http://localhost:3000/"]
             interval: 15s
             timeout: 10s
             retries: 3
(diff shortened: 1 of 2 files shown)


mxkaske:
@claude review

PR #2366, opened 2026-07-10
feat: send OTel metrics from private-location checkers

thibaultleouay:
## What

Private-location checkers (`apps/checker` in private mode) run monitor checks locally but **never emitted OpenTelemetry metrics** — the public checker already does. This wires up **HTTP and TCP** private checks to record the same OTel metrics to the customer-configured OTLP endpoint.

The per-monitor OTel config (`otel_endpoint` / `otel_headers`) already lived in the DB; the missing pieces were propagating it (plus a region label) from `apps/private-location` to the checker over the Connect RPC, and calling the existing `pkg/otel` recorders from the checker's private jobs.

## Changes

**Proto** (`packages/proto/internal/private_location/v1/`)
- New `otel.proto` with a shared `OtelConfig { endpoint, repeated Headers }` (the `Headers` message moved here so HTTP + TCP can both reference `OtelConfig` without an import cycle).
- `otel_config` added to `HTTPMonitor` + `TCPMonitor`; `region` added to `MonitorsResponse`. `DNSMonitor` untouched.
- Regenerated Go into both `apps/checker/proto` and `apps/private-location/proto`.

**Private-location server** (`apps/private-location`)
- `Monitors` looks up the location by token and returns `region = private_location.name`.
- Selects `otel_endpoint` / `otel_headers`, maps them to `OtelConfig` (nil when no endpoint), parsing headers via exported `ParseOtelHeaders`.
- Invalid-token behavior preserved (`sql.ErrNoRows` → empty response, not an error).

**Checker** (`apps/checker`)
- `JobRunner.HTTPJob` / `TCPJob` take a `region`; scheduler threads `MonitorsResponse.region` in.
- `HTTPJob` / `TCPJob` populate `req.OtelConfig` and call `otel.RecordHTTPMetrics` / `RecordTCPMetrics` after the retry loop. On a hard failure the error is synthesized onto the response so the `openstatus.error` counter fires (parity with the public checker).

## Region label

The `openstatus.probes` attribute is the **location name** (server-provided), so metrics are human-readable in the customer's collector with no operator config. (This differs from the numeric id the Tinybird ingest pipeline uses.)

## Out of scope

- **DNS** — `DNSJob` is currently a no-op, so DNS checks don't run in private locations at all; DNS OTel is blocked on implementing DNS end-to-end.
- **Exporter pooling** — `pkg/otel` still sets up/tears down a MeterProvider per check (unchanged from the public path); a future optimization for tight intervals.

## Testing

- `apps/checker/pkg/job/otel_wiring_test.go` (new): a fake OTLP collector + local target servers assert an export lands on HTTP/TCP success **and** on hard failure (error counter), and that nothing is recorded when no endpoint is set.
- `apps/private-location/internal/server/otel_test.go` (new): `Monitors` returns `region`, maps the OTel config with parsed headers, returns nil config when unset, and `ParseOtelHeaders` handles valid/null/empty/invalid input.
- `go build`, `go vet`, and `go test ./...` are green in both apps.

> Note: a full-deployment run against a live OTLP collector was not performed; the wiring tests drive the real job → `pkg/otel` recorder → OTLP HTTP exporter → collector path as a substitute.

https://claude.ai/code/session_01QyhZyvvNzES3koSoX9rWjh

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2366?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


thibaultleouay:
@claude review this carefully

thibaultleouay:
@claude review this again

thibaultleouay:
@claude  review this again

PR #2395, opened 2026-07-16
test(header-analysis): cover the response header parsers

mariazuheros:
The parsers under `packages/header-analysis/src/parser` had no tests apart from `x-vercel-cache`, so I added one file per parser following the same colocated pattern.

What each one covers:

* `parseCfRay`: resolves the Cloudflare data center from the IATA code, fails with a clear message when the code is not in the list, and fails when the header carries no code at all.
* `parseCfCacheStatus`: the five documented states (HIT, MISS, BYPASS, EXPIRED, DYNAMIC), the case insensitive match, and the fallback for an unknown state.
* `parseXVercelId`: a chain of several regions resolved in order, a single region id, and a header with no region ids.
* `parseFlyRequestId`: the region from the trailing IATA code, an unlisted region, and a header with no code.
* `parseCacheControlHeader`: numeric directive values, directives without a value, several directives with loose spacing, and the case insensitive name match.

All pure functions, so no network or database is involved.

```
cd packages/header-analysis
deno test --parallel -A --no-check --sloppy-imports src/parser/
```

27 steps pass, and `deno check` is clean on each new file.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2395?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/header-analysis/src/parser/cache-control.test.ts
+++ b/packages/header-analysis/src/parser/cache-control.test.ts
@@ -1,0 +1,41 @@
+import { expect } from "@std/expect";
+import { describe, it } from "@std/testing/bdd";
+
+import { parseCacheControlHeader } from "./cache-control";
+
+describe("parseCacheControlHeader", () => {
+  it("reads a numeric directive value and keeps the raw directive", () => {
+    const [maxAge] = parseCacheControlHeader("max-age=3600");
+    expect(maxAge.name).toBe("max-age");
+    expect(maxAge.value).toBe(3600);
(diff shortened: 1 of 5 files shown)


thibaultleouay:
@claude review this

PR #2397, opened 2026-07-16
fix: prevent Resend 409 idempotency conflicts on subscriber notifications

mxkaske:
## Summary

Fixes [OPENSTATUS-7PQ](https://openstatus.sentry.io/issues/OPENSTATUS-7PQ): `POST /api/trpc/lambda/subscriberNotification.statusReport` failed with a Resend `409 invalid_idempotent_request`, silently dropping subscriber emails for that status-report update.

**Root cause:** the email channel derives its Resend idempotency key from the status-report update id alone (`status-report-update:<updateId>:<chunkIndex>`). Resend remembers a key for 24 hours and rejects reuse with a *different* request body — which happens when the same update is dispatched twice but the payload changed in between (subscriber list gained/lost a confirmed subscriber, or the report was edited before a re-trigger). The `EmailClient` retry loop then retried the deterministic 409 three more times before swallowing it via `console.error` (the log line Sentry captured).

## Changes

- **`packages/subscriptions/src/channels/email.ts`** — include an FNV-1a fingerprint of the payload (recipients + tokens, page identity, title, status, message, date, components) in the idempotency key. Identical retries still dedupe — the key's actual job, protecting the in-process retry loop from double-sending — while a changed payload gets a fresh key and the emails go out instead of 409ing. FNV-1a keeps the channel Edge-safe (no `node:crypto`).
- **`packages/emails/src/client.tsx`** — `sendStatusReportUpdate` / `sendMaintenanceNotification` no longer retry deterministic Resend rejections (`invalid_idempotent_request`, `invalid_idempotency_key`, `validation_error`); transient errors keep the full 3-retry exponential backoff.
- Test updates: key format/stability/variance cases in `email.test.ts`, and a no-retry-on-409 case in `client.test.ts`.

## Verification

jsr.io is network-blocked in the sandbox so the deno suites couldn't run there — please let CI confirm them. Verified standalone against the pinned versions: `effect@3.21.2` retry semantics (transient→ok = 2 calls, 409 = 1 call, always-transient = 4 calls) and fingerprint determinism/variance; typechecked the new `while:` retry option against real `effect` + `resend@6.12.3` types with `tsc --strict`.

Fixes OPENSTATUS-7PQ

https://claude.ai/code/session_01QJfEKSKiBASmqXcb6TgPKD

---
_Generated by [Claude Code](https://claude.ai/code/session_01QJfEKSKiBASmqXcb6TgPKD)_

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2397?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/emails/src/client.test.ts
+++ b/packages/emails/src/client.test.ts
@@ -39,6 +39,11 @@ function baseReq(
 const ok = { data: { data: [] }, error: null } as any;
 // biome-ignore lint/suspicious/noExplicitAny: simulated Resend application error
 const fail = { data: null, error: { name: "application_error" } } as any;
+// biome-ignore lint/suspicious/noExplicitAny: simulated Resend 409 conflict
+const idempotencyConflict = {
+  data: null,
+  error: { name: "invalid_idempotent_request" },
+} as any;
 
 describe("EmailClient.sendStatusReportUpdate - idempotency & chunking", () => {
   let client: EmailClient;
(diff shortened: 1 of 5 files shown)


mxkaske:
@claude review

PR #2399, opened 2026-07-19
fix(checker): apply header assertion result to request status for private locations

### LINKED ISSUE #2374: Header assertions are being ignored for private location scheduler
## Description

Based on the checker logic below, all requests that fails header assertions will not update the success status to false. As a result, requests will still be wrongly classified as a success:

https://github.com/openstatusHQ/openstatus/blob/15244795aec0e9ad9d5ef2a08a922ca6ba1c723e/apps/checker/pkg/job/http_job.go#L161

arhxam:
## Problem

For monitors running on **private locations**, header assertions are evaluated but their result is ignored, so a request that fails a header assertion is still classified as a **success**.

Fixes #2374

## Root cause

In the private-location scheduler (`apps/checker/pkg/job/http_job.go`), the header-assertion loop called `HeaderEvaluate` but threw the boolean result away:

```go
assert.HeaderEvaluate(string(headersAsString))
```

The status-code and body assertion loops right below it correctly fold the result into `isSuccessful` (`isSuccessful = isSuccessful && ...`), and so does the public checker (`apps/checker/handlers/checker.go:351`). Only the header path in the private-location job dropped it.

## Fix

Combine the header-assertion result into `isSuccessful`, matching the sibling assertions and the public checker:

```go
isSuccessful = isSuccessful && assert.HeaderEvaluate(string(headersAsString))
```

## How I verified

Added `TestHTTPJob_HeaderAssertions` in `apps/checker/pkg/job/http_job_test.go`, which spins up an `httptest` server returning a known header and checks both directions:

- a **failing** header assertion → `RequestStatus == "error"` (`Error == 1`)
- a **passing** header assertion → `RequestStatus == "success"` (`Error == 0`)

I confirmed the test **fails on `main`** (the failing assertion was reported as a success) and **passes with the fix**. `go test ./pkg/job/ ./pkg/assertions/` and `go vet ./pkg/job/` are green.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/checker/pkg/job/http_job.go
+++ b/apps/checker/pkg/job/http_job.go
@@ -158,7 +158,7 @@ func (jr jobRunner) HTTPJob(ctx context.Context, monitor *v1.HTTPMonitor, region
 					Target:     assertion.Target,
 					Key:        assertion.Key,
 				}
-				assert.HeaderEvaluate(string(headersAsString))
+				isSuccessful = isSuccessful && assert.HeaderEvaluate(string(headersAsString))
 			}
 		}
 
--- a/apps/checker/pkg/job/http_job_test.go
+++ b/apps/checker/pkg/job/http_job_test.go
@@ -2,6 +2,8 @@ package job_test
 
 import (
 	"context"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/openstatushq/openstatus/apps/checker/pkg/job"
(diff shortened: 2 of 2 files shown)


thibaultleouay:
@claude review this

PR #2401, opened 2026-07-20
test(tracker): cover the date helpers and blacklist lookup

mariazuheros:
The helpers in `packages/tracker/src` had no tests, so I added them.

* `startOfDay` and `endOfDay`: return midnight and the last millisecond of the day in UTC, and do not mutate the input.
* `isSameDay`: true for two times on the same UTC day, false across a UTC midnight boundary.
* `isInBlacklist`: returns the stored reason for a blacklisted day and undefined for a normal one.

Everything is UTC based, so the tests are deterministic regardless of the local timezone.

```
cd packages/tracker
deno test --parallel -A --no-check --sloppy-imports src/utils.test.ts src/blacklist.test.ts
```

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/tracker/src/blacklist.test.ts
+++ b/packages/tracker/src/blacklist.test.ts
@@ -1,0 +1,15 @@
+import { expect } from "@std/expect";
+import { describe, it } from "@std/testing/bdd";
+
+import { blacklistDates, isInBlacklist } from "./blacklist";
+
+describe("isInBlacklist", () => {
+  it("returns the stored reason for a blacklisted day", () => {
+    const key = "Fri Aug 25 2023";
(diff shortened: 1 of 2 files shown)


mxkaske:
this is legacy and should not be used anymore

PR #2427, opened 2026-07-24
test(services): cover redactSensitiveHeaders

mariazuheros:
The header redactor used before response logs are stored had no tests, so I added them.

`redactSensitiveHeaders` covers:

* `null` input returns an empty object;
* headers matched by an exact sensitive name (authorization, cookie, x-api-key) are replaced with the redacted marker;
* the name match is case insensitive;
* headers whose name contains a sensitive part (session, token, secret, credential) are also redacted;
* non-sensitive headers keep their original values.

Pure function, no database needed.

```
cd packages/services
deno test --no-check --sloppy-imports -A src/monitor/__tests__/response-logs-internal.test.ts
```

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2427?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/services/src/monitor/__tests__/response-logs-internal.test.ts
+++ b/packages/services/src/monitor/__tests__/response-logs-internal.test.ts
@@ -1,0 +1,58 @@
+import { expect } from "@std/expect";
+import { describe, it } from "@std/testing/bdd";
+
+import { redactSensitiveHeaders } from "../response-logs-internal";
+
+describe("redactSensitiveHeaders", () => {
+  it("returns an empty object for null", () => {
+    expect(redactSensitiveHeaders(null)).toEqual({});
+  });
+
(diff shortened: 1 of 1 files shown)


mxkaske:
we already have `response-logs.test.ts` where we test `redactSensitiveHeaders` - feel free to extend that file!

mxkaske:
feel free to reopen another pr but include the tests in `response-logs.test.ts`

PR #2442, opened 2026-07-25
fix: prevent input text from overlapping copy button in dark mode

### LINKED ISSUE #2441: BUG: Text is visible through copy button in dark mode on status page
On the status page, the "Get Updates" popup has a button in each tab, where the text is visible(see through) with the button to copy the link in input in dark mode.
(See the attached image below)

Image

rushil-b-patel:
Cause: The button's outline variant uses dark:bg-input/30
which resolves to a semi-transparent background in dark mode,
letting input text show through.

Fix: Add backdrop blur to the button.

| Before | After |
| -------- | -------- |
| image | image |

Fixes: #2441 

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2442?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/components/blocks/status-updates.tsx
+++ b/packages/ui/src/components/blocks/status-updates.tsx
@@ -126,7 +126,7 @@ export function StatusUpdatesCopyInput({
             withToast: true,
           })
         }
-        className="absolute top-1/2 right-2 size-6 -translate-y-1/2"
+        className="absolute top-1/2 right-2 size-6 -translate-y-1/2 backdrop-blur-xs"
       >
         {isCopied ? <Check /> : <Copy />}
         <span className="sr-only">{labels.ariaCopyLink}</span>


rushil-b-patel:
Hello @thibaultleouay,
I still think that adding `padding-right` on input is clean solution than doing this.

What's your view on it ?

thibaultleouay:
Let's add @mxkaske 🤗

mxkaske:
@rushil-b-patel good catch on that, its not very nicely done! 💯

How about adding an `backdrop-blur` (e.g. `backdrop-blur-xs`) to it?


See
CleanShot 2026-07-26 at 10 54 24@2x

rushil-b-patel:
Hello @mxkaske, With backdrop blur the text would still be behind the button. Don't you think we should avoid the text overflow under the button instead of hiding that overflow behind the button ?

After all its your decision : )

aggmoulik (approved):

mxkaske:
> Hello @mxkaske, With backdrop blur the text would still be behind the button. Don't you think we should avoid the text overflow under the button instead of hiding that overflow behind the button ?
> 
> After all its your decision : )

Lets use backdrop blur @rushil-b-patel. Adding the padding cuts off the text. With the blur, you know there is something happening.

mxkaske:
merci!

PR #2527, opened 2026-08-02
fix: correct monitor sorting by group and groupOrder on /monitors page

### LINKED ISSUE #2526: Status page: Monitors are not ordered the same way as they are on the status page
Public monitors on the status page appear to be ordered based on component groups, but not ordered correctly within the component groups.
This means that a component group with "Service 1, Service 2, Service 3" might be ordered as "Service 2, Service 1, Service 3" depending on their global order.

zapteryx:
The /monitors page was displaying monitors without proper sorting, causing incorrect ordering within component groups. Monitors were only filtered by visibility but not sorted.

This fix implements proper sorting logic that:
1. Groups monitors by their monitorGroupId
2. Orders groups by their position (using minimum order of monitors in group)
3. Sorts monitors within each group by their groupOrder field

This matches the sorting behavior used in the trackers logic on the main status page and ensures monitors appear in the correct order as configured in the dashboard.

Resolves #2526 

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2527?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/status-page/src/app/(status-page)/[domain]/[locale]/(public)/monitors/page.tsx
+++ b/apps/status-page/src/app/(status-page)/[domain]/[locale]/(public)/monitors/page.tsx
@@ -34,6 +34,17 @@ export default function Page() {
 
   if (!page) return null;
 
+  // Precompute group minimum order map for O(n log n) sorting performance
+  const groupMinOrderMap = new Map<number, number>();
+  for (const monitor of page.monitors) {
+    const groupId = monitor.monitorGroupId;
+    if (groupId !== null) {
+      const order = monitor.order ?? 0;
+      const currentMin = groupMinOrderMap.get(groupId) ?? Number.MAX_SAFE_INTEGER;
+      groupMinOrderMap.set(groupId, Math.min(currentMin, order));
+    }
+  }
+
   const publicMonitors = page.monitors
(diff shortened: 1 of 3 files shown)


thibaultleouay:
@claude  review this

apps/status-page/src/app/(status-page)/[domain]/[locale]/(public)/monitors/page.tsx
```diff
@@ -34,7 +34,50 @@ export default function Page() {
 
   if (!page) return null;
 
-  const publicMonitors = page.monitors.filter((monitor) => monitor.public);
+  // Precompute group minimum order map for O(n log n) sorting performance
+  const groupMinOrderMap = new Map<number, number>();
```
Comment (mxkaske) [inline apps/status-page/src/app/(status-page)/[domain]/[locale]/(public)/monitors/page.tsx]: @zapteryx I'd argue we can move that to the server (`trpc.statusPage.get` or the service function so the client doesn't have to bother about it, and we can add a test)

PR #2534, opened 2026-08-03
fix(server): include region in raw check timings

Kathircpe:
## Summary
The POST /v1/check/http response now includes region in each entry of the raw timings array, so consumers can tell which region each timing belongs to.

Previously only timing was returned, dropping the region the checker had already reported (ResponseSchema.region).

## Notes
Additive, backward-compatible API change. No DB migration.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2534?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/server/src/routes/v1/check/http/post.test.ts
+++ b/apps/server/src/routes/v1/check/http/post.test.ts
@@ -52,6 +52,7 @@ test("Create a single check  ", async () => {
         tlsHandshakeStart: 5,
         transferDone: 10,
         transferStart: 9,
+        region: "ams",
       },
     ],
     response: {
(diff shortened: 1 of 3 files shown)


thibaultleouay:
@Kathircpe  thanks for the pr but we are deprecating v1 and focusing on the proto api

Kathircpe:
@thibaultleouay Thanks for the review — understood that v1 is being deprecated and the team is focused on the proto/Connect API. I'll close this PR rather than invest in a surface that's winding down.

I'd love to keep contributing. Could you point me at any open tasks on the proto API (or anything else that would be helpful)? Happy to pick up something.

PR #2541, opened 2026-08-04
perf: dedupe and batch the dashboard shell queries

mxkaske:
The dashboard shell batch was ~7 procedures and ~20 Turso round-trips. Every Drizzle statement is a separate HTTP request on the libsql http driver, so round-trips dominate the p95 rather than query cost. This takes the batch to ~8 round-trips, and the deepest procedure from 5 serial hops to 3.

Follows `85d39deed` (per-procedure `trpc.timing`) and `aedde07ce` (resolve the active workspace once per request), which made the intra-batch split measurable in the first place.

| change | before | after |
| --- | --- | --- |
| `user.get` | 1 query (row already in `ctx.user`) | 0 |
| `workspace.list` | 1 query — the same join `resolveActiveWorkspace` had just run | 0 — served from `ctx.workspaces` |
| `workspace.get` | relational read materializing every page + component + monitor (config blobs) + notification to call `.length` | 6 statements, 1 batched round-trip, index-only counts |
| `page.list` | 2 serial: pages, then every status report of every page | 1 batched round-trip: pages + a distinct unresolved-report probe |
| `monitorTag.list` | tags + join rows via `with: { monitor: true }` — a second copy of the monitor table, unread by any consumer | 1 flat select |
| `statusReport.list` enrich | 4 serial hops (updates → impacts → assocs → pages) | 2 — only impacts depends on a prior result |
| `monitor.list` / `statusReport.list` | `count(*)` always run, `totalSize` discarded by both routers | derived from a short page; only runs when a full page comes back |

### Supporting change

`batchReads(db, [...])` in `packages/services/src/context.ts` — one libsql round-trip for independent reads, degrading to `Promise.all` inside an interactive transaction (no batch endpoint there; the session is baton-chained).

### Wire change

`PageListItem.statusReports: StatusReport[]` → `hasActiveStatusReport: boolean`. The sidebar dot in `nav-status-pages.tsx` was the only reader, and the full report history it fetched is unbounded.

### Notes

- Test suites are included for each touched service (`context`, monitor, monitor-tag, page, status-report, workspace); they need `turso dev` plus a seeded db to run.
- Out of scope, still open from the perf investigation: Tinybird `revalidate: 0` in prod, the duplicate http/tcp `globalMetrics` / `workspace30d` hook pairs, and the mutation paths (`monitor.delete`, `pageComponent.updateOrder`) at p95 ~12s.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2541?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/dashboard/src/components/nav/nav-status-pages.tsx
+++ b/apps/dashboard/src/components/nav/nav-status-pages.tsx
@@ -119,10 +119,6 @@ export function NavStatusPages() {
                 toast.success("Status Page ID copied to clipboard");
               },
             });
-            const hasActiveStatusReport = item.statusReports.some(
-              (report) => report.status !== "resolved",
-            );
-
             return (
               <SidebarMenuItem key={item.id}>
                 <SidebarMenuButton
@@ -150,7 +146,9 @@ export function NavStatusPages() {
                       className={cn(
                         "absolute top-1/2 left-1/2 h-2 w-2 -translate-x-1/2 -translate-y-1/2 rounded-full",
                         STATUS[
-                          hasActiveStatusReport ? "degraded" : "operational"
+                          item.hasActiveStatusReport
+                            ? "degraded"
+                            : "operational"
                         ],
                       )}
                     />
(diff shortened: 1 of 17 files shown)


mxkaske:
@claude review

PR #2542, opened 2026-08-04
chore: split workspace usage from get, tidy dashboard tRPC setup

mxkaske:
Two small cleanups on top of the dashboard shell prefetch work.

### Split `usage` out of `workspace.get`

`workspace.get` was calling `getWorkspaceWithUsage`, so every route in the shell paid for five `count(*)` queries plus a redundant re-select of the workspace row — even though the authed middleware has already resolved and parsed that row, and only two surfaces (billing, onboarding checklist) actually read the counts.

- `workspace.get` now returns `ctx.workspace` directly — no query.
- New `workspace.usage` procedure backed by `getWorkspaceUsage`, which returns just the counts (`WorkspaceWithUsage` is gone; `WorkspaceUsage` stays).
- Consumers that only needed `limits` read them off `workspace.get`; billing and the nav checklist opt into `workspace.usage`.

The count queries themselves are unchanged (active monitors only, everything workspace-scoped), so the existing `getWorkspaceUsage` tests carry over with the result shape flattened from `result.usage.x` to `result.x`.

### tRPC client/server tidy-up

- `apps/dashboard/src/lib/trpc/client.tsx` had its own inline `makeQueryClient` that drifted from the shared `query-client.ts` one — it was missing the `dehydrate.shouldDehydrateQuery` option that keeps pending queries streamable. No behaviour change today (the browser instance is never dehydrated), but it means future global defaults land in one place instead of two.
- `fetchQueryOrNotFound` matched `error instanceof TRPCClientError`, which fails across bundle boundaries and never matches the bare `TRPCError` an in-process caller throws. Replaced with a duck-typed `isNotFoundError` check on `code` / `data.code`, the same fields tRPC matches on internally.

<sub>Both commits are dashboard-only apart from the `workspace` service and router.</sub>

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2542?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/dashboard/src/app/(dashboard)/layout.tsx
+++ b/apps/dashboard/src/app/(dashboard)/layout.tsx
@@ -50,9 +50,6 @@ async function HydrateSidebar({ children }: { children: React.ReactNode }) {
     queryClient.prefetchQuery(trpc.page.list.queryOptions()),
     queryClient.prefetchQuery(trpc.monitor.list.queryOptions()),
     queryClient.prefetchQuery(trpc.workspace.get.queryOptions()),
-    // The sidebar's getting-started banner reads these. Awaited only because
-    // nothing streams yet — it moves to `batchPrefetch` with step 2b.
-    queryClient.prefetchQuery(trpc.workspace.usage.queryOptions()),
     queryClient.prefetchQuery(trpc.workspace.list.queryOptions()),
     queryClient.prefetchQuery(trpc.user.get.queryOptions()),
   ]);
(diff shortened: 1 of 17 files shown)


mxkaske:
@claude review

PR #2547, opened 2026-08-04
slack: use assistant.threads.setStatus instead of a "Thinking..." message

thibaultleouay:
The agent posted a real ":hourglass_flowing_sand: Thinking..." message and
mutated it via chat.update into the answer, the confirmation card, or an
error. That placeholder notifies, shows up in thread history (and had to be
filtered back out before feeding the model), and leaves a stale "Thinking..."
artifact when a run dies mid-flight.

Slack's assistant.threads.setStatus renders an ephemeral loading indicator
instead. Since the March 2026 scope change it accepts chat:write, which the
app already has, so it works in channels with no manifest or scope change.

Dropping the placeholder means the reply is a fresh post rather than an
update, so:

- postThreadReply centralizes the cannot_reply_to_message top-level fallback,
  now shared by the answer, confirmation and error paths.
- setStatus is best-effort — failures are logged and the run continues, since
  a cosmetic indicator must never silence the bot. A 60s refresh keeps it
  alive past Slack's ~2 minute drop, and it is cleared before the reply lands.
- A second destructive draft in a thread now strips the old card's buttons and
  marks it superseded, rather than leaving two identical live cards sharing
  one action id.
- store() takes a caller-supplied id so the card's blocks can be built before
  the message exists; replace() also updates the stored messageTs.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_018vZGKg3npYcKypYaYEDkJr

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2547?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/server/src/libs/test/doubles/slack-test-state.ts
+++ b/apps/server/src/libs/test/doubles/slack-test-state.ts
@@ -10,6 +10,7 @@ export interface SlackTestState {
   postMessageOverride: Override;
   updateOverride: Override;
   postEphemeralOverride: Override;
+  setStatusOverride: Override;
   runAgentOverride: (() => Promise<unknown>) | null;
   repliesImpl: () => Promise<unknown>;
 }
@@ -22,6 +23,7 @@ if (!g.__slackTestState) {
     postMessageOverride: null,
     updateOverride: null,
     postEphemeralOverride: null,
+    setStatusOverride: null,
     runAgentOverride: null,
     repliesImpl: () =>
       Promise.resolve({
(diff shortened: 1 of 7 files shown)


thibaultleouay:
@claude review this

PR #2550, opened 2026-08-05
perf: improve db queries

mxkaske:
<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2550?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Stops the dashboard's RSC render from blocking on queries it doesn't need. Deletions only, no new loading states.

### Three changes

**1. Prefetches moved to the route that actually reads them.**
`notification.list` sat on the shared `monitors/[id]/layout.tsx`, but its only consumer is `forms/monitor/update.tsx` — the edit form. Overview, logs and incidents were all paying for a query they never read. Moved down into `monitors/[id]/edit/layout.tsx`.

**2. Prefetches of sidebar-owned queries removed.**
`monitor.list` and `page.list` are prefetched by the root `(dashboard)/layout.tsx` for the sidebar, and the sidebar (`nav-monitors.tsx`, `nav-status-pages.tsx`) is mounted on every dashboard route and never unmounts — so its observer keeps both permanently in the client cache. Every route-level prefetch of them was redundant:

- On a **hard load**, `getQueryClient` is React-cached per request, so the root resolves them first and the child prefetch hits a fresh entry (`staleTime: 60s`) without fetching. Already a no-op.
- On a **client navigation**, the root layout doesn't re-execute, so the child prefetch issued a real DB query — while the client already held the data.

Removed from `monitors/(list)/page.tsx`, `status-pages/(list)/page.tsx`, `status-pages/[id]/layout.tsx`, `status-pages/[id]/components/layout.tsx`, `status-pages/[id]/history/layout.tsx`, `overview/layout.tsx`.

**3. Sequential awaits collapsed.**
`monitors/[id]/layout.tsx` awaited the `monitor.get` 404 gate and *then* a `Promise.all`. Nothing depends on the gate's result, so both now start in the same tick — one wave instead of two.

### Why `monitor.list` in particular

It's the expensive one. Unlike `page.list`, it doesn't use `batchReads` (`packages/services/src/context.ts:90`, which collapses reads into a single libSQL `db.batch()`): it runs a select plus four `Promise.all` queries as separate round trips. That made it the tail of every parallel batch it sat in, so removing it moves wall-clock rather than just deleting a line.

### Effect

- **Hard loads: unchanged.** The removed prefetches were already no-ops there.
- **Client navigation: fewer blocking queries.** `/status-pages` and `/status-pages/[id]/edit` now do zero DB work in the RSC render; the others lose the slowest member of their parallel batch.
- **Monitor detail:** one round trip saved per navigation.

### Notes for review

- Every consumer of the removed queries uses plain `useQuery` — nothing suspends, so no Suspense boundary is needed anywhere.
- **Behavior change:** on a client nav where the cached copy is older than the 60s `staleTime`, you now see cached rows immediately with a background refetch, instead of server-fresh rows swapped in by hydration. No spinner, no skeleton — but a record renamed in another tab can show its old value briefly. Consistent with how `staleTime` already behaves elsewhere.
- **Not measured.** The round-trip counts come from reading the service implementations, not from timings.


mxkaske:
@claude review

PR #2562, opened 2026-08-08
docs: server and db scripts run on Deno, not Bun

no-hup:
Two small corrections after the Bun-to-Deno migration:

- `apps/server/README.md` lists Bun under Tech, but the server runs entirely on Deno now — every script in `apps/server/package.json` is `deno run`/`deno test`, and the Dockerfile builds from `denoland/deno` and ships a `deno compile` binary.
- `apps/dashboard/README.md` says Bun runs the db env/migrate/seed scripts, but all of `packages/db`'s scripts are `deno run` too.

Left the root README's Bun requirement alone since Bun is still genuinely used there (root `env` script and `apps/screenshot-service`).

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2562?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ ghcr.io/openstatushq/openstatus-checker:latest
 - [Node.js](https://nodejs.org/en/) >= 20.0.0
 - [pnpm](https://pnpm.io/) >= 10.26.0
 - [Bun](https://bun.sh/)
+- [Deno](https://deno.com/)
 - [Turso CLI](https://docs.turso.tech/quickstart).
 
 #### Dashboard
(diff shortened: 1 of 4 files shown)


no-hup:
Pushed a follow-up after a second look: added Deno to the root README requirements (it's now needed for the db scripts and server but wasn't listed), fixed `apps/workflows/README.md` which still said `bun install`/`bun run dev` while its scripts are all Deno, and kept an optional Bun note in the dashboard README since `src/scripts` utilities still use it.

mxkaske:
thanks!

PR #2564, opened 2026-08-09
fix: api server key scope

mxkaske:
## Summary

API keys carry `read` or `write` scopes. In affected versions, scope enforcement was applied only to the v1 REST API (`/v1/*`, via the `requireWriteScope` middleware) and to mutations routed through `@openstatus/services`, which call `requireScope`. Several Connect RPC handlers under `/rpc` instead performed database writes with inline queries and never checked the caller's scope.

A user holding a `read`-scoped API key for a workspace could therefore create, update and delete monitors, status-page components and component groups, unsubscribe status-page subscribers, and trigger on-demand monitor runs — the last of which also dispatches outbound HTTP requests from openstatus infrastructure and consumes the workspace's synthetic-check quota.

## Impact

Privilege escalation within an already-authenticated workspace. The affected calls remain workspace-scoped, so this does not grant cross-workspace or cross-tenant access, and it does not expose data a read key could not already read.

## Patches

All affected RPC handlers now route their mutations through `@openstatus/services`, where `requireScope(ctx, "write")` runs before any database write or outbound request.

---

reported by **Jaejun Cho** via email **Private Security Disclosure: Read-Only API Key Can Create Persistent HTTP Monitors**

### DIFF
--- a/apps/server/src/routes/rpc/handlers/monitor/converters/assertions.test.ts
+++ b/apps/server/src/routes/rpc/handlers/monitor/converters/assertions.test.ts
@@ -1,0 +1,204 @@
+import { assertion } from "@openstatus/assertions";
+import {
+  type BodyAssertion,
+  type HeaderAssertion,
+  NumberComparator,
+  type RecordAssertion,
+  RecordComparator,
+  type StatusCodeAssertion,
+  StringComparator,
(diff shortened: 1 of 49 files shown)


mxkaske:
@claude review

mxkaske:
@claude review

PR #2569, opened 2026-08-09
chore: improve repo agent experience

mxkaske:
Makes the repo — rather than a maintainer — catch what's wrong, so the core team moves faster and outside contributions become safe to accept.

The bottleneck for an agent working in this repo is the feedback loop, not reading comprehension. More prose doesn't move it; a fast scoped verify command and machine-checkable conventions do. Ordering: **feedback loop → enforcement → knowledge.**

## Verify loop

```sh
pnpm verify        # oxfmt + oxlint + doc refs + deno check. No database, ~9s cold.
pnpm verify:test   # tests for the packages your diff touches.
```

`pnpm check` was **failing locally while passing in CI**, in 15 packages, all with `Cannot find name 'process'` / `'Buffer'`. It was the toolchain: deno only injects node globals when a `node:` specifier is present until 2.9, and `devbox.json` had no deno at all while `deno-version: v2.x` floated CI up to 2.9.5.

So `devbox.json` now pins one exact set — `nodejs@24.12.0`, `deno@2.9.4`, `turso-cli@1.0.15`, `bun@1.3.13`, `sqld@0.24.32` — and every workflow pins the same node and deno. With that, `pnpm check` is 39/39 locally, including inside `devbox run`, which previously couldn't run it at all. A verify command that goes red for reasons unrelated to your change is one agents learn to ignore.

## CI gates

- **Blocking `pnpm lint` in `check.yml`.** `lint.yml` is autofix.ci — it commits fixes, it does not gate. Nothing was enforcing lint before this.
- **`pnpm check:docs`** (`scripts/check-doc-refs.mts`, node type-stripping, no deps): every `docs/…` citation in the repo and every backticked repo path inside an `AGENTS.md` must resolve. It immediately found four source files citing `docs/adr/test-audit-cleanup.md`, **a file that has never existed in git history**. Those now point at `packages/services/AGENTS.md`, which carries the invariant they were reaching for. ADR 0008 also got its missing index row.
- **The PR review bot can now run commands.** `claude-code-review.yml` had exactly two steps — checkout and the action — so every review opened with "I couldn't run `tsc`/`oxlint` in this sandbox". It now gets `check.yml`'s setup preamble and `--allowed-tools` for the verify commands. Both Claude workflows are gated on `github.repository_owner`, so fork PRs don't get an unfixable red X.

## Context files

`AGENTS.md` at the root (90 lines, cross-cutting truth only) plus seven nested files: `packages/services`, `apps/{dashboard,server,status-page,workflows,checker,web}`. `CLAUDE.md` becomes a plain-prose pointer — no vendor `@import` syntax, so every tool can read it. Its stale four-step test ritual is deleted; `apps/dashboard/README.md` already documents the correct path.

`docs/adr/` is **demoted, not deleted**. 6 of the 9 ADRs were written on one day as a backfill; since then, 3 ADRs in 265 commits, and the index step already failed on the most recent one. They're also the wrong shape — ADR 0008 says "`loadSeededWorkspace` is gone", which is a diff, not a state. The set is now marked frozen, the "when to write an ADR" obligation is dropped, and the load-bearing facts are restated in present tense in the right `AGENTS.md`.

## Services triage + enforcement

`scripts/oxlint-plugin-openstatus.js` adds two rules via `jsPlugins`:

- **`services-mutation-guards`** (error) — a function in `packages/services` that calls `withTransaction` must also call `requireScope` and `emitAudit`. Keyed on behaviour, not filename.
- **`no-db-barrel-in-client`** (error) — a `"use client"` file must not value-import `@openstatus/db/src/schema`; it drags drizzle and the whole schema graph into the browser bundle.

Plus `no-restricted-imports` banning `node:*` in `packages/services`, which is reachable from Edge and from Deno.

Gaps were fixed before the rules landed, so the exemption list encodes no bugs:

- `monitor/relations.ts` — three verbs mutating monitor regions, tags and notifiers with **no `requireScope`**. Reached today only from tRPC (user actors, where the check is a no-op), so nothing was exploitable, but an MCP or API-key caller would have been.
- `import/run.ts` — `runImport` creates monitors and pages with no scope check. A read-only key could run an import.
- `packages/db/src/schema/status_reports/constants.ts` — new drizzle-free module so client forms can import `statusReportStatus` without the barrel (4 call sites fixed).

Twelve sites are genuine exemptions and carry an inline disable with the reason above it — subscriber-actor verbs where `requireScope` is a documented no-op, chat sessions (per-user UI state, no audit action exists), and two verbs that delegate their audit row to a shared internal helper.

Worth noting: `member/delete.ts` and `workspace/downgrade.ts` **look** like missing-audit bugs by grep and are not. The first audits inside `removeMemberInWorkspace`; the second routes every trim step through an already-audited entity verb.

## Housekeeping

`ralph/` deleted (unused). `CONTRIBUTING.MD` → `CONTRIBUTING.md`, pointing at the verify loop.

## Verification

`pnpm verify` green. All six test legs pass individually: services 73, api 35, server 236, workflows 7, subscriptions 9, rest 25/25 packages.

Both new lint rules were negative-tested — reintroduce a violation and they fire; remove it and they go quiet.

## Follow-ups, not in this PR

- **Turn on required status checks.** `main` has branch protection but the API returns no `required_status_checks` key at all, so `Check`, `Tests` and the aggregate `🧪 Tests` job are all advisory. An error-severity rule in a job nobody must pass changes nothing.
  - `Check` is safe to require now.
  - `Tests` is too: I triaged all 3 failures in the last 30 runs on `main` and **none is a merged regression**. 08-02 was `SQLITE_BUSY: database is locked`; 08-05 and 08-06 were `apps/server` rpc suites colliding on shared seeded fixtures (`pageId 1 does not match the page (7) of the selected components` — the code raised a correct error for the state it saw). Those suites still load the seeded page instead of minting a workspace, which is residual ADR-0008 debt worth its own PR.
- Dropping DB-backed tests into the review sandbox would need an `sqld` service container plus migrate/seed, at roughly 2 min on every PR. The no-database `pnpm verify` tier works without it.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mxkaske:
@claude review

PR #2594, opened 2026-08-20
fix(status-page): keep slug prefix in links on subdomain-shaped hosts

thisisharsh7:
This PR fixes the broken navigation on themes.openstatus.dev, where the status
page links dropped the page prefix.

## Before

Clicking Status, Events or Monitors led to a 404, because the prefix was decided
from the hostname shape alone and that host does not own a page.

https://github.com/user-attachments/assets/909d6a09-5c56-4a9e-ab81-4190065b2221

## After

The path is checked too and wins when it already carries the page, so the links
resolve correctly. Status pages on their own subdomain or custom domain are
unaffected.

https://github.com/user-attachments/assets/ad9f00c2-1ae7-41e7-a922-20f0c9e70f52

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


thisisharsh7:
I checked this. /status gives a 404 already, so no links are shown there and nothing breaks.

https://themes.openstatus.dev/status -> 404
https://themes.openstatus.dev/status/en -> works

Also removing the locale check would break pages whose name is the same as a page like "monitors".

mxkaske:
@thisisharsh7 thanks for raising the point! two things:

1. can you include unit tests in `resolve-pathname-prefix.test.ts` so that we won't break it accidentally again
2. can we restrict the slug to `status` when the domain is `themes.openstatus.dev`?

thisisharsh7:
@mxkaske both done added unit tests, and the prefix now only applies on the theme explorer host with the status slug.

mxkaske:
@claude review

mxkaske:
@thisisharsh7 see the reviews from claude and cubic - lets make it restrict to the canonical `themes.openstatus.dev` 🙏

thisisharsh7:
thanks already pushed, it matches only `themes.openstatus.dev` now

mxkaske:
@claude review

PR #2605, opened 2026-08-26
fix: correct jobType guard in DNS monitor update route

### LINKED ISSUE #2604: PUT /v1/monitor/dns/{id} returns 404 for every DNS monitor
Updating a DNS monitor through the v1 API always fails with a 404, even when the monitor exists and belongs to the workspace.

Repro against the public API:

1. `POST /v1/monitor/dns` with a valid body — returns 200, monitor created with `jobType: "dns"`.
2. `PUT /v1/monitor/dns/{id}` with the same body — returns `404 Monitor {id} not found`.

Cause: the jobType guard in `apps/server/src/routes/v1/monitors/put_dns.ts` checks `_monitor.jobType !== "tcp"` (looks copy-pasted from `put_tcp.ts`), so a real DNS monitor never passes it. Side effect: a TCP monitor *can* be updated through the DNS route, which then overwrites its url/timeout with DNS-shaped data.

The fix is the one-character change to `!== "dns"`, matching the guards in `put_http.ts` and `put_tcp.ts`. Happy to open a PR with the fix plus tests for both cases.

no-hup:
Fixes #2604

`put_dns.ts` guarded on `jobType !== "tcp"`, so `PUT /v1/monitor/dns/{id}` 404'd for every DNS monitor, while TCP monitors could be updated through the DNS route. Changed the guard to `!== "dns"`, matching `put_http.ts` / `put_tcp.ts`.

Added two tests to `put_dns.test.ts`: create a DNS monitor via the API, update it, expect 200; and updating a TCP monitor via the DNS route now returns 404. Both create their own fixtures and clean up after themselves — the shared seed is untouched.

Full `apps/server` suite passes locally against a fresh sqld: 247 passed (1021 steps), 0 failed.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/server/src/routes/v1/monitors/put_dns.test.ts
+++ b/apps/server/src/routes/v1/monitors/put_dns.test.ts
@@ -3,6 +3,113 @@ import { test } from "@std/testing/bdd";
 
 import { app } from "@/index";
 
+import { MonitorSchema } from "./schema";
+
+test("update a dns monitor", async () => {
+  const create = await app.request("/v1/monitor/dns", {
+    method: "POST",
+    headers: {
+      "x-openstatus-key": "1",
+      "content-type": "application/json",
+    },
+    body: JSON.stringify({
+      frequency: "10m",
+      name: "OpenStatus",
+      description: "OpenStatus website",
+      regions: ["ams", "gru"],
+      request: {
+        uri: "openstatus.dev",
+      },
+      active: true,
+      public: true,
+    }),
+  });
+
+  expect(create.status).toBe(200);
+  const created = MonitorSchema.parse(await create.json());
+
+  const res = await app.request(`/v1/monitor/dns/${created.id}`, {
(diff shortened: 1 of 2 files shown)


thibaultleouay:
v1 is legacy we are not updating it

no-hup:
Fair enough, good to know v1 is frozen — I'll leave it as is. The issue is there in case anyone hits the 404 and goes looking. Thanks for the quick look.

PR #2613, opened 2026-08-26
feat: gRPC health check monitors

thibaultleouay:
Adds a fifth monitor type. A gRPC monitor calls `grpc.health.v1.Health/Check` on the target and reports the serving status it answers with — the same check Kubernetes, Envoy and `grpcurl` speak, so most services need no new endpoint.

**The probing tier and the TypeScript tier move together in this PR**, per `apps/checker/AGENTS.md`: the Go checker, the private-location agent, the proto contracts, Tinybird, and the product surface are all here.

## Shape

| Server answer | Result | `completed` |
| --- | --- | --- |
| `SERVING` | success (degraded past the threshold) | yes |
| `NOT_SERVING` | error — "service reports NOT_SERVING" | yes |
| `SERVICE_UNKNOWN` | error — names the service | yes |
| `UNIMPLEMENTED` | error — "server does not implement grpc.health.v1.Health" | yes |
| `DEADLINE_EXCEEDED` / `UNAVAILABLE` / TLS failure | error | no |

`completed` is the axis the whole feature hangs off. It cannot be derived from the error flag — `NOT_SERVING` is a failed check that completed perfectly well — so `GRPCResult` carries it explicitly, and four behaviours read it: retry, OTel branching, what `latency`/`timing` hold, and the latency-quantile filter.

## Decisions worth reviewing

- **Health protocol only.** No reflection, no arbitrary unary calls.
- **Two additive `monitor` columns** (`grpc_service`, `grpc_tls`) rather than encoding config into the URI. A URI parser would have had to exist in Go twice and TypeScript four times — the cross-boundary duplication `AGENTS.md` calls a defect.
- **Only transport failures are retried.** A definitive answer is recorded once, the way an HTTP assertion failure already is (`handlers/checker.go` returns `nil` after one). Retrying `NOT_SERVING` three times only delays the alert by the backoff window.
- **`service` and `tls_mode` are `optional`** in the proto, so they use `!== undefined` rather than a sentinel. An empty service clears it back to overall-health; a patch that omits `tls_mode` leaves a plaintext monitor plaintext instead of silently switching it to TLS.
- **Timings reuse HTTP's ten-field shape verbatim**, so `calculateTiming` and the dashboard waterfall work unchanged.
- **`UNIMPLEMENTED` gets its own message.** The server is reachable, it just never registered a health service; reading that as "down" sends people hunting the wrong problem.
- **TLS failures return a fixed string.** Go's raw error quotes the peer's certificate subject and chain, which must not reach the caller.
- **`uri` is validated as `host:port`** by a `buf.validate` pattern, mirrored in `validateMonitorPatchBounds` — the dashboard's pre-create Test gate lives in the tRPC router and never runs for a monitor created over ConnectRPC.

## Not inherited from the ICMP pipes

The three defects still live in the ICMP datafiles are fixed in the gRPC copies: no `SELECT *` in a materialization, the `regions` parameter is actually used in `by_interval`, and the 30d/90d latency endpoints honour the forwarded window. Latency quantiles additionally filter on `servingStatus IS NOT NULL`, so a refused connection no longer drags the median toward zero — `quantileIf` in the mixed pipes, because a `WHERE` there would have silently filtered the `countIf` counts too.

## Two cleanups in passing

- Adding the fifth ingest validator was the moment to extract the monitor-id / latency / timestamp checks the four existing ones duplicated into one helper.
- `mapMonitors` gained a logging `default:` — it had no fallthrough, so an unsupported job type was dropped with no row, no log and no error.
- `data-table-basics.tsx` went from 846 lines with four near-identical components to 684 with shared row primitives and five composed ones.

## Verification

- `pnpm verify` — 39/39
- `go test ./...` in `apps/checker` and `apps/private-location` — green, race-clean
- `buf lint api` — clean
- **`tb --cloud deploy --check` — "Deployment is valid"** against the real workspace
- New tests: real in-process health servers covering all six outcomes and all three TLS modes (including that a rejected self-signed cert never leaks its subject), retry-count assertions, timestamp stamping on both job paths, ingest range checks, and the RPC create/update surface.

`pnpm verify:test` could **not** be run to completion: every DB-backed suite hangs on this machine's shared local libSQL, including files this PR never touches — `packages/services/.../workspace.test.ts` and the `main` version of `.../monitor.test.ts` both hang identically. That is the condition `AGENTS.md` describes; it needs a reseed. Everything not needing the database passes.

## Rollout

Deploy order matters: **promote the Tinybird deployment before deploying the checker fleet**, and note the promotion is the one step a revert cannot undo. Mixed-version safety is symmetric during the rolling window — an old checker against a new private-location schedules no gRPC jobs, and a new checker against an old private-location gets an empty `grpc_monitors` list and never calls `IngestGRPC`.

Worth asking polylane for the checker fleet's headroom first: a gRPC check holds an HTTP/2 connection and a TLS handshake, heavier per check than an ICMP echo on 512 MB VMs.

https://claude.ai/code/session_01YDMfxo4TidoFQt93wcnsYm

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2613?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


thibaultleouay:
@claude review this carefully

thibaultleouay:
@claude review this again

thibaultleouay:
@claude review this again

PR #2616, opened 2026-08-27
fix: derive status bar radius from --radius

mxkaske:
Lets themes pick a real border radius for the status bar instead of the current all-or-nothing pill/square, without adding a theme variable.

### Problem

`.rounded-full` is globally redefined as `calc(var(--radius) * 99999999)` in `apps/status-page/src/app/globals.css` and `apps/dashboard/src/app/globals.css`. That multiplier is what makes the status bar binary: `--radius: 0` renders square, any non-zero value renders a full pill. A theme that sets `--radius: 0.25rem` cannot get 4px corners on the bar.

The status bar is not conceptually a pill. It is a rounded rect that happens to be narrower than twice its radius, so `rounded-full` was a misuse.

### Fix

Use `--radius` directly on the bar. CSS clamps `border-radius` proportionally when the radii on a side exceed that side's length, and the bars are always narrow — `max-w-2xl` with 45 days (`fillStatusDataFor45Days`) and `gap-px` gives roughly 13px per bar — so the clamp produces the pill for free:

| theme | `--radius` | renders | change |
| --- | --- | --- | --- |
| Openstatus | `0rem` | square | none |
| Openstatus (Rounded) | `0.625rem` = 10px | clamps to 6.7px = full pill | none |
| Gruvbox, Tomorrow | `0.25rem` = 4px | real 4px corners | new |

No new CSS variable, `THEME_VAR_NAMES` untouched.

### Changes

- `packages/ui/src/components/blocks/status-bar.tsx`: `rounded-full` to `rounded-(--radius)` on the item wrapper and the clipping container; per-segment `rounded-t-full`/`rounded-b-full` dropped, since the parent's `overflow-hidden` already clips to the radius.
- Both `globals.css` files: removed the now-dead `.rounded-t-full` / `.rounded-b-full` overrides. The status bar was their only consumer. `.rounded-full` stays in both, since the event dots, status icon and badges still rely on it to square off under `--radius: 0`.

### Notes

- Gruvbox and Tomorrow already ship `--radius: 0.25rem`, so they change appearance too, not just future themes. That is the intended fix, but it is three themes shifting rather than one.
- On narrow mobile viewports (~6px bars) a 4px radius still clamps back to a pill. Inherent to clamping, and arguably the right behaviour.
- Overlaps with #2598, which solves the same problem by adding a `--status-bar-radius` theme variable. If this lands, that PR can drop its `--status-bar-radius` and `status-bar.tsx` changes and ship the Probo palette on its own.

Verified: `tsc --noEmit` and `oxlint` clean in `packages/ui`.

https://claude.ai/code/session_0131zqiLo8GntEFdxkq7Fz5n

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2616?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mxkaske:
@claude review

PR #2620, opened 2026-08-30
fix: update Hono security patches

yoominho91:
## Summary

- update the Hono catalog entry from 4.12.21 to 4.12.34
- override transitive Hono copies so the MCP SDK cannot retain 4.12.21
- pin `@hono/node-server` to the patched 1.19.15 release

## Security impact

The server currently uses `cors()` with default options in:

- `apps/server/src/routes/public/index.ts`
- `apps/server/src/routes/v1/index.ts`

Hono 4.12.21 reaches the vulnerable default `allowHeaders` path described in [GHSA-8j4g-w8fx-2239](https://github.com/advisories/GHSA-8j4g-w8fx-2239), where an unauthenticated preflight request can trigger quadratic header parsing and consume disproportionate CPU.

The lockfile also retained vulnerable Hono and Node adapter copies through `@modelcontextprotocol/sdk`. The overrides keep all workspace and transitive copies on patched releases.

## Verification

- `pnpm audit --prod`: Hono / `@hono/node-server` advisories **15 → 0** (2 high, 12 moderate, 1 low removed)
- `pnpm install --frozen-lockfile --lockfile-only`
- `pnpm verify` with the repository-pinned Node 24.12.0 and Deno 2.9.4: **39/39 packages passed**

The lockfile change is intentional and contains only the Hono resolution and peer snapshot updates required by these overrides.


yoominho91:
CI note: the first run passed the API job, while the server job failed during test cleanup with `SQLITE_BUSY: database is locked`. I refreshed the branch because external contributors cannot use GitHub's rerun action. On the second run the server job passed, while the API job hit the same `SQLITE_BUSY` error in unrelated database tests.

All other GitHub checks passed, and the failure moved between unrelated suites without a code change. Could a maintainer rerun the failed Tests workflow? The three Vercel statuses also report `Authorization required to deploy`, which requires team approval.

yoominho91:
Addressed in 1feeecd: the Hono override now uses `hono: "catalog:"`, so the catalog is the single version source while transitive Hono copies remain pinned. I also tested the suggested `$hono` form with pnpm 11.2.1, but it fails here because the workspace root has no direct `hono` dependency. Frozen lockfile install, the targeted Hono audit (0 advisories), and `pnpm verify` (39/39) all pass with the catalog override.

thibaultleouay:
@claude review this

yoominho91:
Friendly ping — the review you kicked off came back with no correctness or security issues, cubic's catalog/override note is addressed in `1feeecd`, and all checks are green. Nothing outstanding on my side.

Anything you'd like changed before this goes in?

PR #2621, opened 2026-08-30
fix(self-hosting): run database migrations in Coolify deployment

### LINKED ISSUE #2532: Self-hosting: coolify-deployment.yaml has no db-migrate step, leaving the database unmigrated
## Summary

Following `COOLIFY_ENVIRONMENT_GUIDE.md` produces an instance that starts (8/8 containers healthy) but is **unusable**: the libSQL database is never migrated, so sign-in fails with `LibsqlError` on the `user` table.

The root cause is that **`coolify-deployment.yaml` does not include a `db-migrate` service**, unlike `docker-compose-lightweight.yaml` which does. And the image that service relies on is **not publicly available**, so there is no documented way to run the migrations manually.

## Steps to reproduce

1. In Coolify, create a Docker Compose service and import `coolify-deployment.yaml` as instructed by the guide
2. Set the required variables from the guide (`DATABASE_URL=http://libsql:8080`, `DATABASE_AUTH_TOKEN=`, `AUTH_SECRET`, `NEXT_PUBLIC_URL`, `SELF_HOST=true`, `RESEND_API_KEY`)
3. Deploy, then open the dashboard and attempt to sign in

## Expected

The database schema is created and sign-in works.

## Actual

All 8 containers report healthy, the dashboard renders its sign-in page, but authentication fails:

```
[auth][error] AdapterError
[auth][cause]: Error: Failed query: select "id", "tenant_id", ... from "user" where "user"."email" = ?
  "name": "LibsqlError"
```

The database is empty — confirmed directly against libSQL:

```bash
curl -X POST http://libsql:8080/v2/pipeline \
  -H "Content-Type: application/json" \
  -d '{"requests":[{"type":"execute","stmt":{"sql":"SELECT name FROM sqlite_master WHERE type=\"table\""}},{"type":"close"}]}'
# => "rows": []
```

Attempts to run the migration image manually all fail:

```
docker pull openstatus/db-migrate:latest
# => pull access denied, repository does not exist or may require 'docker login'

docker pull ghcr.io/openstatushq/openstatus-db-migrate:latest
# => error from registry: denied
```

The `server` and `workflows` images are distroless (no shell), and no migration files could be found inside any running container, so there is no in-place workaround either.

## Suggested fix

Either add a `db-migrate` service to `coolify-deployment.yaml` (as in `docker-compose-lightweight.yaml`) **and** publish the image, or document how to run migrations against a self-hosted libSQL instance.

## Two other issues found along the way

**1. `workflows` requires GCP, though the guide lists it as optional.** The container crash-loops on startup without `GCP_PROJECT_ID` and `GCP_LOCATION`, because it initialises a Cloud Tasks client eagerly:

```
TypeError: render fails for not matching
    at PathTemplate.render (google-gax/build/src/pathTemplate.js:124:27)
    at CloudTasksClient.queuePath (@google-cloud/tasks/build/esm/src/v2/cloud_tasks_client.js:1243:53)
```

Since the five remaining services `depends_on` it, the whole stack stays in `Created`. Setting dummy values unblocks startup, but it is unclear whether scheduled checks work without a real Cloud Tasks queue — worth clarifying in the guide.

**2. Env validation requires SaaS-only variables.** `dashboard` and `status-page` both refuse to boot with `Invalid environment variables` unless these are set, even with `SELF_HOST=true`:

`STRIPE_SECRET_KEY`, `PROJECT_ID_VERCEL`, `TEAM_ID_VERCEL`, `VERCEL_AUTH_BEARER_TOKEN`, `UNKEY_API_ID`, `UNKEY_TOKEN`, `TINY_BIRD_API_KEY`

None of these are relevant to a self-hosted instance, and none are mentioned in the guide. Making them optional when `SELF_HOST=true` would remove a fair amount of friction.

## Environment

- Coolify 4.1.2, Ubuntu 24.04, Docker 29.7.1
- `coolify-deployment.yaml` from `main`, imported as documented
- All images pulled at their `latest` tag

solomonerous:
Fixes #2532

`coolify-deployment.yaml` starts all eight containers but never migrates libSQL, so a fresh deploy looks healthy and then sign-in fails with `LibsqlError` on the `user` table. Three changes:

1. `docker-publish.yml` now builds the existing one-shot migration runner (`packages/db/Dockerfile`) and publishes it as `ghcr.io/openstatushq/openstatus-db-migrate` (added to the dispatch default, the push-detection list, and the build matrix).
2. `coolify-deployment.yaml` gains a `db-migrate` service - idempotent drizzle run, `restart: on-failure:3`, `pull_policy: always` so redeploys pick up new migrations - and `workflows`, `server`, `dashboard` and `status-page` now wait on `service_completed_successfully`.
3. `COOLIFY_ENVIRONMENT_GUIDE.md` says migrations run automatically on every deploy.

The approach copies what `docker-compose-lightweight.yaml` and `docker-compose.github-packages.yaml` already do (the latter even carries a comment saying "a ghcr-published image could replace it later"). `checker` and `private-location` chain off `server` health, so they inherit the ordering without changes.

One caveat: the published image only shows up on GHCR once this reaches `main`, so I couldn't run a real Coolify import against it beforehand. Both YAML files parse, and `pnpm verify` passes 39/39.

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -55,7 +55,7 @@ jobs:
             if echo "$CHANGED" | grep -q '^packages/'; then
               SERVICES_LIST=("server" "dashboard" "workflows" "private-location" "status-page" "checker" "db-migrate")
             else
-              for svc in server dashboard workflows private-location status-page checker db-migrate; do
+              for svc in server dashboard workflows private-location status-page checker; do
                 if echo "$CHANGED" | grep -q "^apps/$svc/"; then
                   SERVICES_LIST+=("$svc")
                 fi
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -18,7 +18,7 @@ on:
       services:
         description: 'Services to build (comma-separated)'
         required: false
-        default: 'server,dashboard,workflows,private-location,status-page,checker'
+        default: 'server,dashboard,workflows,private-location,status-page,checker,db-migrate'
         type: string
 
 concurrency:
@@ -53,9 +53,9 @@ jobs:
 
             # packages/** changes affect all services
             if echo "$CHANGED" | grep -q '^packages/'; then
-              SERVICES_LIST=("server" "dashboard" "workflows" "private-location" "status-page" "checker")
+              SERVICES_LIST=("server" "dashboard" "workflows" "private-location" "status-page" "checker" "db-migrate")
             else
-              for svc in server dashboard workflows private-location status-page checker; do
+              for svc in server dashboard workflows private-location status-page checker db-migrate; do
                 if echo "$CHANGED" | grep -q "^apps/$svc/"; then
                   SERVICES_LIST+=("$svc")
                 fi
@@ -101,6 +101,9 @@ jobs:
           - service: checker
             context: apps/checker
             dockerfile: apps/checker/Dockerfile
+          - service: db-migrate
+            context: .
+            dockerfile: packages/db/Dockerfile
 
     steps:
       - name: Checkout repository
--- a/COOLIFY_ENVIRONMENT_GUIDE.md
+++ b/COOLIFY_ENVIRONMENT_GUIDE.md
@@ -12,6 +12,8 @@ This guide explains how to configure environment variables for OpenStatus deploy
    ```
 3. Click **"Deploy"**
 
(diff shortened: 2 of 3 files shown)


thibaultleouay:
@claude review this

PR #2618, opened 2026-08-28
fix(support): reject whitespace-only input and show a type error

thisisharsh7:
This PR fixes three validation issues in the Support dialog: the name and message fields accepted whitespace-only input, and an unselected Type surfaced a raw Zod error string.

**Before**

https://github.com/user-attachments/assets/6b3cd5e0-0a03-4e72-977f-3d81f3f6eda2

**After**

https://github.com/user-attachments/assets/1ddaefc0-224a-4a35-9c64-83dd95622351

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2618?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2632, opened 2026-09-03
fix(validation): reject whitespace-only names across dashboard forms

thisisharsh7:
This PR fixes eight forms that accepted a whitespace-only name or title. In each case the client schema used `z.string().min(1)` without `.trim()`, so a value of only spaces passed validation and was either saved as a blank-looking record or rejected by the server with a raw Zod error string shown to the user.

### 1. API key name

Submitting spaces sent an empty name to the server, which surfaced the raw `ZodError` JSON in a toast. Now the field blocks it inline and no request is sent.

**Before**

01-api-key

**After**

01-api-key

---

### 2. Private location name

The location saved with a blank Name, leaving an unidentifiable row in the table.

**Before**

02-private-location

**After**

02-private-location

---

### 3. Monitor tag name

The tag saved as "Tags saved successfully" with an empty label, and stayed empty after a reload.

**Before**

03-monitor-tag

**After**

03-monitor-tag

__

> Note: monitor-tag and page-component updateOrder previously had no length constraint, so this adds one. a workspace with a pre-existing blank tag or component name will need to rename it on the next save.

---

### 4. Status report title

The report was created and listed with an empty Title column.

**Before**

04-status-report

**After**

04-status-report

---

### 5. Maintenance title

The maintenance saved with an empty Title while the message was kept, so the list row had nothing to identify it by.

**Before**

05-maintenance

**After**

05-maintenance

---

### 6. Page component name

The component name was replaced with spaces and saved, blanking the label on the status page preview.

**Before**

06-page-component

**After**

06-page-component
---

### 7. Feedback message

The widget accepted spaces and sent them to `feedback.submit`, which already trims server-side (#2618), so the request always failed with a raw error. The client schema was the missing half of that fix — the message is now blocked before any request is made.

**Before**

https://github.com/user-attachments/assets/b6506212-45a8-48a7-8bfa-4e4b6ca9a37e

---

### 8. Onboarding component name

A whitespace-only component name was silently discarded: the status page published successfully but the component was never created, leaving the new workspace with "No components selected".

**Before**

08-onboarding

**After**

08-onboarding

<a href="https://cubic.dev/pr/openstatusHQ/openstatus/pull/2632?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/dashboard/src/components/forms/api-key/form.tsx
+++ b/apps/dashboard/src/components/forms/api-key/form.tsx
@@ -34,7 +34,7 @@ import { toast } from "sonner";
 import { z } from "zod";
 
 export const schema = z.object({
-  name: z.string().min(1, "Name is required"),
+  name: z.string().trim().min(1, "Name is required"),
   description: z.string().optional(),
   expiresAt: z.string().optional(),
   // Single-value radio. The wire format on the create-key API is
--- a/apps/dashboard/src/components/forms/components/form-components.tsx
+++ b/apps/dashboard/src/components/forms/components/form-components.tsx
@@ -106,7 +106,7 @@ const componentSchema = z.object({
   id: z.number(),
   monitorId: z.number().nullish(),
   order: z.number(),
-  name: z.string().min(1, { message: "Name is required" }),
+  name: z.string().trim().min(1, { message: "Name is required" }),
   description: z.string().optional(),
   type: z.enum(["monitor", "static"]),
 });
--- a/apps/dashboard/src/components/forms/components/form-components.tsx
+++ b/apps/dashboard/src/components/forms/components/form-components.tsx
@@ -117,7 +117,7 @@ const schema = z.object({
     z.object({
       id: z.number(),
       order: z.number(),
-      name: z.string(),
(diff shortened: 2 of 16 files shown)


thibaultleouay:
@claude review

thisisharsh7:
fixes two thing.

1. claude review that group name should also not be empty so added that validation to the input 
2. I was seeing that the error name is required is overflowing the container so styled it accordingly



before -


image


after -

image

thisisharsh7:
another fix I found while working on the review that workspace name can still be empty so I fixed that as well.


before -


https://github.com/user-attachments/assets/4b1cab53-50e2-47e1-a64f-726c17ab84e6




after -


image

mxkaske:
dope! thanks @thisisharsh7 🚀

